PR #6, opened 2018-07-12
add template for .circleci config for node libraries

crwilcox:


setup.py
```diff
@@ -47,5 +47,10 @@
     packages=packages,
     install_requires=dependencies,
     include_package_data=True,
+    package_data={
+        'synthtool': [
+            'gcp/templates/**/*',
+            'gcp/templates/node_library/.circleci/*'],
```
crwilcox: dot files aren't being taken without being specific

synthtool/transforms.py
```diff
@@ -80,7 +80,7 @@ def _copy_dir_to_existing_dir(
         excludes = []
     for root, _, files in os.walk(source):
         for name in files:
-            rel_path = str(Path(root).relative_to(source)).lstrip(".")
```
crwilcox: this was causing issues with hidden dirs/files. I didn't see adverse effects from removing this, but @theacodes maybe you recall why we did this?

PR #14, opened 2018-07-18
add public flag to circle config publish

crwilcox:


kinwa91:
I think it should be `npm publish --access=public`. I opened a PR #17 without realizing there's one created already :)

Closing in favour of #17.

PR #18, opened 2018-07-19
Update README to explain how to use templating.

crwilcox:


README.md
```diff
@@ -9,9 +9,11 @@ pyenv to get 3.6.
...
 
 ```
-git clone sso://devrel/cloud/libraries/tools/synthtool
-cd synthtool
-pip install .
+# Install latest
+pip install git+https://github.com/GoogleCloudPlatform/synthtool.git
+
+# Install stable
+pip install gcp-synthtool
```
JustinBeckwith: I am python stupid.  For whatever reason, I had to do something like:

```
python3 -m pip install --user --upgrade gcp-synthtool
```

For a typical setup, will `pip` refer to python2 or 3? What about the `--user` and `--upgrade` flags?

README.md
```diff
@@ -41,7 +43,21 @@ Once you run synthtool without errors:
...
-- TODO. Should be similar to Python
+Node.js generation is similar to python. For an example `synth.py` you can look at https://github.com/googleapis/nodejs-speech/blob/master/synth.py.
 
 ### Ruby
 - TODO. Should be similar to Python
+
+
+## Features
+### Templating
+Synthtool supports template files using jinja. As an example let's look at node.js. The templates for node can be found at `/synthtool/gcp/templates/node_library/`. As of writing this doc, the following works in a node.js synth script.
```
JustinBeckwith: Lets leave out the `As of writing this doc`.  Hope is that we keep the README up to date as we make changes :)

README.md
crwilcox: On linux systems python/pip refer to python2, at least typically. Python 3 is usually pip3 and python3. that said, in my user path pip and python both refer to python3. So This isn't a hard and fast rule.

As far as the difference they are the same, though I am going to change it to what you wrote. pip is an executable and module. the way you wrote it loads pip from that specific python, so there is a bit better chance of it not getting crossed up :)

--user refers to not trying to place the package in the user folder (think inverse of the -g flag for npm)

--upgrade tells it to upgrade and not just say 'hey you have this package'

PR #24, opened 2018-08-01
Use new Nox

theacodes:
Change-Id: Ibcde6d08ddd057db62b1bc1e5354b7f39dab1594

### DIFF
--- a/nox.py
+++ b/nox.py
@@ -15,16 +15,14 @@
 import nox
 
 
-@nox.session
+@nox.session(python='3.6')
 def blacken(session):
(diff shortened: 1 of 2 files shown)


nox.py
```diff
@@ -15,26 +15,23 @@
 import nox
 
 
-@nox.session
+@nox.session(python='3.6')
```
crwilcox: should we rev to 3.7 . at the same time?

theacodes: Not yet, I wanna cut over all of our tools to 3.7 when I get back from vacation.

PR #27, opened 2018-08-06
Repository generator

chingor13:
This PR adds a simple generator class for checking out a git repository and executing commands within the temporary directory.

This will enable synth scripts for languages that use a custom generator (like Discovery-based generators).


theacodes:
@chingor13 can you help me understand how this is different from the `git` source [here](https://github.com/GoogleCloudPlatform/synthtool/blob/master/synthtool/sources/git.py)? I think I'm missing something.

chingor13:
I guess we're not *really* adding any logic here -just abstracting behind a "generator" class. Example [here](https://github.com/chingor13/google-api-php-client-services/blob/92855f5b7a9a3c107a2fb220d85ecbab7b593f47/synth.py).

I'll try it with the git and shell commands directly and see how it looks.

theacodes:
Cool, let me know how it goes. Maybe we can just augment the git source, but totally open to a higher-level wrapper if needed.

chingor13:
Here's an [example using the synthtool git/shell directly](https://github.com/chingor13/google-api-php-client-services/blob/6158d9055762c68eceab4c60366462ccf3bb6cca/synth.py). The questionable thing is using `_tracked_paths` which looks like its an internal structure, but is necessary for doing the copying.

theacodes:
Cool, let's just get the git one to add any cloned repos to tracked paths.

chingor13:
Closed in favor of using git/shell commands directly in #31

PR #31, opened 2018-08-08
Add git sources to tracked_paths

chingor13:
Automatically track all cloned git repo paths. Many code generators operate within their own repository and do not provide an option for where the generated code is placed. For these synth scripts, the user can use the `git` and `shell` tools to run the codegen process. The output files need to be in a tracked path or they cannot be found and the user should not be setting `_tracked_paths` themselves.

This PR makes #27 unnecessary.

### DIFF
--- a/synthtool/sources/git.py
+++ b/synthtool/sources/git.py
@@ -15,6 +15,7 @@
 import pathlib
 import shutil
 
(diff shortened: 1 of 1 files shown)


theacodes:
Thanks, @chingor13!

PR #34, opened 2018-08-17
feat: (nodejs) read repo_name and package_name from package.json

kinwa91:


synthtool/sources/node_package.py
```diff
@@ -0,0 +1,37 @@
...
+    read package name and repository in package.json from a Node library.
+
+    Returns:
+        package_name: Name of the package (with scope)
+        repo_name: Name of the repository
+    """
+    with open('./package.json') as f:
+        data = json.load(f)
+
+        package_name = data['name']
```
theacodes: Instead of plucking values, why not just pull the whole contents of package.json?

synthtool/gcp/common.py
```diff
@@ -27,9 +27,10 @@ def __init__(self):
     def py_library(self) -> Path:
         raise NotImplemented()
 
-    def node_library(self, package_name, repo_name) -> Path:
+    def node_library(self, **kwargs) -> Path:
+        metadata = node_package.read_metadata()
         t = templates.TemplateGroup(_TEMPLATES_DIR / "node_library")
-        result = t.render(package_name=package_name, repo_name=repo_name)
+        result = t.render(**{**metadata, **kwargs})
```
theacodes: neat, but let's go with:

```
metadata = ...
metadata.update(kwargs)

t.render(**metdata)
```

or even better:

```
kwargs['metadata'] = node_package.read_metadata()

t.render(**kwargs)
```

synthtool/sources/node_package.py
kinwa91: Probably won't need all those fields, also helps to verify those fields exists first.

theacodes: you can still do that without plucking the fields. Just `if not 'name' in data` etc. If you go ahead and pull the full metadata, we'll save ourselves from having to do a bunch of little change/releases over time to add more fields if we ever need them.

synthtool/sources/node_package.py
```diff
@@ -0,0 +1,35 @@
+# Copyright 2018 Google LLC
```
theacodes: Can you follow the convention set by #22 and use `languages/nodejs`?

PR #36, opened 2018-08-21
Fix tracked paths issue by preferentially matching the longest path

theacodes:


theacodes:
> Can we add a test for this expected behavior?

Done.

PR #47, opened 2018-08-28
Preserve file mode in templates

### LINKED ISSUE #46: Need templates to carry over file permissions as well
Node.js templates have kokoro configs that need to be executable, but we currently can't keep the +x permission when copying over the templates to the destination.

kinwa91:
Fixes #46. Pardon my novice python, any edits welcome :)


theacodes:
LGTM, but you'll need to fix the tests (& it would be great if you added new ones specifically for this). @crwilcox to do the final review while I'm OOO.

PR #48, opened 2018-08-29
Move: excludes arg paths should be relative to source, not destination

kinwa91:
```python
transforms.move(
    source,
    destination,
    excludes=['garbage.md']) # after this change, this will be relative to the source, not destination
```

So if there exists a `source/garbage.md`, it will be ignored.

Please let me know if this change makes sense to you.


kinwa91:
This is needed for https://github.com/googleapis/nodejs-firestore/blob/master/synth.py

firestore moved their source location to under dev/

PR #51, opened 2018-08-31
Run `npm install` verbose

JustinBeckwith:


JustinBeckwith:
FYI @soldair

soldair:
could we try to replace `npm install --verbose` with 
`npx gist:45d18dfcca291e019ef420af7831f96c 10000 2 install`?

alexander-fenster:
Can we try in one repo first?  Go ahead and add it here https://github.com/googleapis/nodejs-kms/tree/npm-verbose just to verify it works, then I think we're good to go.

alexander-fenster:
Actually I will just try :)

PR #63, opened 2018-09-11
nodejs: add test_project arg for templates

kinwa91:
...to specify the project to use when running system/samples-tests


theacodes:
I am *extremely* surprised that this is exposed/hardcoded at all. Can we just treat this like a secret and have kokoro/circle inject the environment variable?

kinwa91:
Let's just go with hard-coding the default here - it's not trivial to add the environment to all nodejs- repositories on Circle and it's even more effort for Kokoro with GCS or Keystore strategies.

theacodes:
@kinwa91 if you still need this, can you resolve the conflicts?

kinwa91:
I think this is ready @theacodes

PR #65, opened 2018-09-11
Correct a wrong path in the PHP template

tmatsuo:


tmatsuo:
cc @jdpedrie @dwsupplee

PR #83, opened 2018-09-25
feat(node): match different styles of repository string

kinwa91:


theacodes:
@kinwa91 there's some lint errors here, do you mind addressing?

PR #90, opened 2018-10-05
Add discogapic generator

chingor13:
Java Compute uses a discogapic generator (the a gapic generator that accepts discovery docs rather than protos). This adds a new generator that fetches the discovery-artifact-manager repo.

The java compute synth would end up like:
```python3
import synthtool as s
import synthtool.gcp as gcp

gapic = gcp.DiscoGAPICGenerator()
common_templates = gcp.CommonTemplates()

library = gapic.java_library(
    service='compute',
    version='v1',
    config_path='artman_compute.yaml',
    artman_output_name='')

s.copy(library / 'gapic-google-cloud-compute-v1/src', 'src')
```

### DIFF
--- a/synthtool/gcp/__init__.py
+++ b/synthtool/gcp/__init__.py
@@ -13,7 +13,9 @@
 # limitations under the License.
 
 from . import gapic_generator
+from . import discogapic_generator
 from . import common
 
+DiscoGAPICGenerator = discogapic_generator.DiscoGAPICGenerator
 GAPICGenerator = gapic_generator.GAPICGenerator
 CommonTemplates = common.CommonTemplates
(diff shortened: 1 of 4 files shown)


chingor13:
I opted for a separate module rather than trying overload the other gapic generator. An alternative would be to add another option to `_generate_code()` for the repo to use. The downside is that it's another repo to sync in the setup.

chingor13:
@theacodes @busunkim96 offered to do the refactor and should be able to push directly here.

chingor13:
Looks like the `git` import was removed in the discogapic generator but is used to checkout the discovery artifact manager repo.

PR #100, opened 2018-10-17
Add option to specify local googleapis directory

### LINKED ISSUE #91: Add support to specify local googleapis dir
Currently , the flow is to checkout googleapis (or googleapis-private) and from there generate. There exist cases where pointing this to a locally modified version would be useful.

busunkim96:
Closes #91.

cc: @beccasaurus

### DIFF
--- a/synthtool/gcp/gapic_generator.py
+++ b/synthtool/gcp/gapic_generator.py
@@ -12,6 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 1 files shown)


busunkim96:
@theacodes So set `self.googleapis` to the value of the `SYNTHTOOL_GOOGLEAPIS` environment variable if it exists? 

Another question: Should the local googleapis be used even when `private=True` in the `_generate_code` call?

PR #101, opened 2018-10-19
docs: update the new issue templates for nodejs

JustinBeckwith:


synthtool/gcp/templates/node_library/.github/bug_report.md
```diff
@@ -0,0 +1,33 @@
+---
+name: Bug report
```
busunkim96: This section gets formatted into a table. Is this intentional? 
7qvs39grae0

JustinBeckwith: Yup!  It should look like this:
https://github.com/googleapis/google-cloud-node/issues/new/choose

PR #116, opened 2018-11-05
Update github issue templates

JustinBeckwith:


.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -0,0 +1,5 @@
+Fixes #<issue_number_goes_here> (it's a good idea to open an issue first for discussion)
```
theacodes: I don't like this template. We have status checks that enforce all of this and it's just noise.

PR #121, opened 2018-11-08
Show stdout for docker pull

### LINKED ISSUE #118: Don't hide docker output
Downloading the `googleapis/artman:` Docker image can take a *long* time.  Beyond the initial "Pulling artman image", synthtool gives feedback on what is occurring, which has led me to Ctrl-C it, assuming that the hub might be down, etc.

Using `stdout=PIPE` swallows output from subprocesses, which is not ideal for something like `docker pull`.

busunkim96:
Closes #118. 

I wasn't sure whether to show `stdout` for _only_ docker pull or every time the `run` function is called.

### DIFF
--- a/synthtool/gcp/artman.py
+++ b/synthtool/gcp/artman.py
@@ -93,4 +93,6 @@ class Artman:
 
     def _install_artman(self):
         log.debug("Pulling artman image.")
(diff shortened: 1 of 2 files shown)


theacodes:
> I wasn't sure whether to show stdout for only docker pull or every time the run function is called.

only for docker pull is fine for now.

PR #129, opened 2018-11-14
Harden replace

### LINKED ISSUE #67: Avoid stomping on read-only files
This usage pattern for fixups:

```python
s.replace(
    '**/*.py',
    <source string>,
    <target string>,
)
```

blows up for unrelated files (e.g., under `.nox/`) which are not writeable (e.g., symlinks into non-local Python installs).

We could:

- Harden `replace` so that it skips over files which aren't writeable.
- Allow the script to pass in a specific `root` to be used for expanding the globs.
- Or maybe document the syntax for globs such that `root` isn't needed?

busunkim96:
Closes #67


synthtool/transforms.py
```diff
@@ -60,9 +61,23 @@ def _expand_paths(paths: ListOfPathsOrStrs, root: PathOrStr = None) -> Iterable[
             )
 
 
+def _file_is_gitignored(path: Path) -> bool:
+    """Checks if a file is ignored by git"""
+    try:
+        subprocess.run(["git", "check-ignore", "-q", str(path.resolve())], stderr=subprocess.PIPE)
```
theacodes: oof, forking for every file is going to cause significant slowdown for synthtool. Can we do this a different way? Do we really need to avoid gitignored files? (I mean, they're ignored anyway).

tseaver: `git check-ignore` can take a list of files on standard input:  it only outputs those which are ignored. See the [manpage](https://git-scm.com/docs/git-check-ignore).

theacodes: That's useful, but I'd still like to understand why we care about synthtool touching ignored files?

theacodes: Lol I realize now it was [my suggestion](https://github.com/googleapis/synthtool/issues/67#issuecomment-420813577). I rescind it. Let's do not that for now.

PR #144, opened 2018-11-20
add noxfile template

crwilcox:


crwilcox:
A result of running templating. https://github.com/googleapis/google-cloud-python/pull/6606

PR #145, opened 2018-11-20
Add ability to format java code with a custom google-java-format version

chingor13:
Usage would be something like:

```python
import synthtool.languages.java as java

# ... do some codegen into ./src
java.format_code("./src")
```

Note the gapic-generator does run version 1.1 of the Google java formatter and this one defaults to 1.6 (most up to date). We can update the one in gapic, but we'd also like the ability to run this on our apiary libraries too.

### DIFF
--- a/synthtool/languages/java.py
+++ b/synthtool/languages/java.py
@@ -39,7 +39,7 @@ def format_code(path: str, version: str = DEFAULT_FORMAT_VERSION) -> None:
 
     # Run the formatter as a jar file
     log.info("Running java formatter on {} files".format(len(files)))
-    shell.run(["java", "-jar", jar, "--replace"] + files)
+    shell.run(["java", "-jar", str(jar), "--replace"] + files)
 
 
 def _download_formatter(version: str, dest: Path) -> None:
--- a/synthtool/languages/java.py
+++ b/synthtool/languages/java.py
@@ -1,0 +1,48 @@
+# Copyright 2018 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+from synthtool import cache
+from synthtool import log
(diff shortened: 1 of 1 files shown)


synthtool/languages/java.py
```diff
@@ -0,0 +1,53 @@
...
+    shell.run(["java", "-jar", jar, "--replace"] + files)
+
+
+def _download_formatter(version: str, dest: Path) -> None:
+    url = f"https://github.com/google/google-java-format/releases/download/google-java-format-{version}/google-java-format-{version}-all-deps.jar"
+    log.info("Downloading java formatter")
+    shell.run(["curl", "-L", "-o", dest, url], hide_output=False)
+
+
+def _find_files(basedir: str, suffix: str = ".java") -> List[str]:
```
theacodes: Can this be replaced with `glob.iglob("**/*.java")`?

synthtool/languages/java.py
```diff
@@ -0,0 +1,53 @@
...
+
+    # Run the formatter as a jar file
+    log.info("Running java formatter on {} files".format(len(files)))
+    shell.run(["java", "-jar", jar, "--replace"] + files)
+
+
+def _download_formatter(version: str, dest: Path) -> None:
+    url = f"https://github.com/google/google-java-format/releases/download/google-java-format-{version}/google-java-format-{version}-all-deps.jar"
+    log.info("Downloading java formatter")
+    shell.run(["curl", "-L", "-o", dest, url], hide_output=False)
```
theacodes: Use Requests, please.

```python
response = requests.get(url)
response.raise_for_status()
with open(dest, "wb") as fh:
    fh.write(response.content)
```

synthtool/languages/java.py
```diff
@@ -0,0 +1,53 @@
...
+    # Find all .java files in path and run the formatter on them
+    files = _find_files(path)
+
+    # Run the formatter as a jar file
+    log.info("Running java formatter on {} files".format(len(files)))
+    shell.run(["java", "-jar", jar, "--replace"] + files)
+
+
+def _download_formatter(version: str, dest: Path) -> None:
+    url = f"https://github.com/google/google-java-format/releases/download/google-java-format-{version}/google-java-format-{version}-all-deps.jar"
```
theacodes: Make this a module-level constant, please.

synthtool/languages/java.py
```diff
@@ -0,0 +1,53 @@
...
+
+import os
+from synthtool import cache
+from synthtool import log
+from synthtool import shell
+from pathlib import Path
+from typing import List
+
+
+def format_code(path: str, version: str = "1.6") -> None:
```
theacodes: Make this default a module-level constant, please.

PR #146, opened 2018-11-22
Update python templates

crwilcox:
Copy black flake8 ignore elements to not conflict with black formatting. 
A good number of our libraries require core as part of their tests. Add this to localdep in noxfile.

### DIFF
--- a/synthtool/gcp/templates/python_library/.flake8
+++ b/synthtool/gcp/templates/python_library/.flake8
@@ -1,4 +1,5 @@
 [flake8]
+ignore = E203, E266, E501, W503
 exclude =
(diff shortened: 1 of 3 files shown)


synthtool/gcp/templates/python_library/.flake8
```diff
@@ -1,4 +1,5 @@
 [flake8]
+ignore = E203, E266, E501, W503
```
tseaver: Ugh, I know I'm late to the party, but it would be way better if this line used named constants. (checks) Double ugh!! `pycodstyle` doesn't *have* named constants!

Could we add a comment there explaining the codes?

PR #159, opened 2018-11-30
fix: pass -md option to openssl

alexander-fenster:
Openssl got upgraded (probably to v1.1.0) in the Docker image used by CircleCI, and it now changed the default algorithm used for decryption, which made all our encrypted keys incompatible. Adding the `-md md5` to `openssl -d` (following [this advice](https://wiki.archlinux.org/index.php/OpenSSL#%22bad_decrypt%22_while_decrypting)) helps ([example workflow](https://circleci.com/workflow-run/2ed13275-1ecf-476d-8644-2435e6394c8b)).


kinwa91:
Let's merge it for now, so at least we can publish in the meantime. @alexander-fenster

alexander-fenster:
OK! Would you mind running autosynth with this change?

PR #169, opened 2018-12-07
Add system tests for gapic

theacodes:


theacodes:
fyi: the PR hasn't been merged because the test requires running docker-in-docker, which doesn't seem to work right because of the mapping needed by artman to access googleapis and the outdir.

It's probably still useful to merge this, but I didn't have time to work through the docker issue.

PR #177, opened 2019-01-09
Add include_protos option to generation

crwilcox:
This option will enable requesting the protos files to be in genfiles.

### DIFF
--- a/synthtool/gcp/discogapic_generator.py
+++ b/synthtool/gcp/discogapic_generator.py
@@ -52,13 +52,7 @@ class DiscoGAPICGenerator:
         return self._generate_code(service, version, "java", **kwargs)
 
     def _generate_code(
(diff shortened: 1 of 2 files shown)


synthtool/gcp/discogapic_generator.py
```diff
@@ -52,7 +52,7 @@ def java_library(self, service: str, version: str, **kwargs) -> Path:
         return self._generate_code(service, version, "java", **kwargs)
 
     def _generate_code(
-        self, service, version, language, config_path=None, artman_output_name=None
+        self, service, version, language, config_path=None, artman_output_name=None, include_protos=False,
```
crwilcox: For python synth.py (the drive to implement this) we use gapic_generator, but I added identical support to discogapic.

PR #187, opened 2019-02-05
node: add lint/fix example to contributing guide

callmehiphop:
Fixes https://github.com/googleapis/nodejs-bigquery/issues/216


JustinBeckwith:
@callmehiphop can you use `github-repo-automation` to craft a PR for this change?  We should land this too, but it's nice to have a PR where we don't need to approve/rename the title 60 times :) Let me know if you need help with that!

PR #202, opened 2019-03-13
Add docs session to noxfile

busunkim96:


busunkim96:
To merge after https://github.com/googleapis/gapic-generator/pull/2619

synthtool/gcp/templates/python_library/noxfile.py.j2
```diff
@@ -140,3 +140,21 @@ def cover(session):
     session.run("coverage", "report", "--show-missing", "--fail-under={{ cov_level if cov_level != None else '100' }}")
 
     session.run("coverage", "erase")
+
+@nox.session(python="3.7")
+def docs(session):
+    """Build the docs."""
```
bcoe: this builds the docs for python libraries specifically? so would this be run as part of the nightly `synthtool` run on `google-cloud-python` as an example?

might be worth fleshing out the comment slightly more to call this out: "generates docs for python libraries nightly"?

will generating docs for other languages require the similar addition of a noxfile?

busunkim96: These templates only get used for the Python monorepo, google-cloud-python. Docs currently get built all together at the top level.

With this addition to the noxfile, you can go to a specific subdirectory like `google-cloud-python/tasks`, run `nox -s docs` and generate documentation for just that package. This is needed for googleapis.dev, which wants docsets by package. It's also faster than building all of the docs.

This won't run in any nightly/release builds right now, but it will be in release builds once Python is ready to move to googleapis.dev

bcoe: :+1: awesome, thanks for clarifying.

PR #208, opened 2019-03-27
chore: improve the README template

bcoe:
organically @callmehiphop has made some great edits to README.md. This modifies our base template to make it lineup.

Now _all_ our libraries can move towards this straight-forward standard.


callmehiphop:
> lets not call Dave out in the commit history

hater!

PR #211, opened 2019-03-27
Check 'metadata' in kwargs.

### LINKED ISSUE #210: Python synth broken
See: https://github.com/googleapis/google-cloud-python/issues/7565

The bug was introduced in PR #206, in [`_generic_library`](https://github.com/googleapis/synthtool/blob/d744f2e261b29e0b22a45c2f9269840edd581972/synthtool/gcp/common.py#L38):

```python
    def _generic_library(self, directory: str, **kwargs) -> Path:
        self._load_generic_metadata(kwargs["metadata"])0
```

That keyword is only present (added explicitly) in `node_library`.  The fix is to use `kwargs.get("metadata", some_appropriate_default)`.

busunkim96:
Fixes #210

### DIFF
--- a/synthtool/gcp/common.py
+++ b/synthtool/gcp/common.py
@@ -35,7 +35,8 @@ class CommonTemplates:
         self._templates = templates.Templates(_TEMPLATES_DIR)
 
     def _generic_library(self, directory: str, **kwargs) -> Path:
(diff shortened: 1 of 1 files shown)


bcoe:
@busunkim96 thank you for addressing this :+1:

PR #213, opened 2019-03-27
feat: skip README rendering if .repo-metadata.json does not exist

bcoe:
This would protected us from the wall of broken README PRs we got last night.

We will want to do the same with `samples/README.md` generation.


JustinBeckwith:
Adding @busunkim96 and @crwilcox because I am barely capable of reading Python let alone review it 😆

Issue #219, opened 2019-04-05
README generation handles acronyms poorly


bcoe:
Noticed in this pull request: https://github.com/googleapis/nodejs-bigquery/pull/411/files

That we handle acronyms poorly during README generation -- perhaps if we see several capital letter consecutively we could skip title-casing and treat it as an acronym.

PR #220, opened 2019-04-09
fix: support consecutive capital letters

bcoe:
The simple algorithm I wrote for decamelize did not support acronyms well (the bug is [seen here](https://github.com/googleapis/nodejs-bigquery/blob/45b6f7634e6517ca217c0d73b59b818256549750/README.md)).

This PR extends the method adding this functionality and adds tests for this edge-case; CC: @steffnay

fixes: #219

### DIFF
--- a/synthtool/__main__.py
+++ b/synthtool/__main__.py
@@ -14,7 +14,7 @@
 
 import os
 import sys
-from importlib import util
+import importlib.util
 from typing import List, Sequence
 
 import click
@@ -70,8 +70,8 @@ def main(synthfile: str, metadata: str, extra_args: Sequence[str]):
     if os.path.lexists(synth_file):
         synthtool.log.debug(f"Executing {synth_file}.")
         # https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly
-        spec = util.spec_from_file_location("synth", synth_file)
-        synth_module = util.module_from_spec(spec)
+        spec = importlib.util.spec_from_file_location("synth", synth_file)
+        synth_module = importlib.util.module_from_spec(spec)
 
         if spec.loader is None:
             raise ImportError("Could not import synth.py")
--- a/synthtool/__main__.py
+++ b/synthtool/__main__.py
@@ -12,9 +12,9 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import importlib
 import os
 import sys
+from importlib import util
 from typing import List, Sequence
 
 import click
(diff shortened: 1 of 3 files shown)


synthtool/__main__.py
```diff
@@ -70,8 +70,8 @@ def main(synthfile: str, metadata: str, extra_args: Sequence[str]):
     if os.path.lexists(synth_file):
         synthtool.log.debug(f"Executing {synth_file}.")
         # https://docs.python.org/3/library/importlib.html#importing-a-source-file-directly
-        spec = importlib.util.spec_from_file_location("synth", synth_file)
-        synth_module = importlib.util.module_from_spec(spec)
+        spec = util.spec_from_file_location("synth", synth_file)
```
bcoe: I was receiving linting errors locally without switching to this approach (upstream regression perhaps?).

synthtool/gcp/common.py
```diff
@@ -153,10 +153,22 @@ def _load_partials(self, metadata: Dict):
 
 def decamelize(str: str):
     """ parser to convert fooBar.js to Foo Bar. """
-    str2 = str[0].upper()
-    for chr in str[1:]:
-        if re.match(r"[A-Z]", chr):
-            str2 += " " + chr.upper()
+    prev = " "
+    str2 = ""
```
tswast: I see this is existing code, but `output` or `decamelized` would be much more readable than `str2`.

synthtool/gcp/common.py
```diff
@@ -153,10 +153,22 @@ def _load_partials(self, metadata: Dict):
...
+    for i in range(0, len(str)):
```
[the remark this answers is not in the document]
tswast: Also, `chr` is a built-in, so it's best to avoid that as a variable name. Maybe `current` to correspond to `prev`?

synthtool/gcp/common.py
```diff
@@ -153,10 +153,22 @@ def _load_partials(self, metadata: Dict):
...
+        chr = str[i]
+        if prev == " ":
+            str2 += chr.upper()
+        elif re.match(r"[A-Z]", chr):
+            # handle "JSONFrom" -> "JSON From" case.
+            next = ""
```
tswast: `next` is also a built-in, so it's best not to overwrite it. Usually we use a trailing underscore to avoid these kinds of collisions. `next_`.

synthtool/gcp/common.py
```diff
@@ -153,10 +153,22 @@ def _load_partials(self, metadata: Dict):
 
 def decamelize(str: str):
```
tswast: `str` is a built-in, so we probably shouldn't use it as a variable name. How about `value` or `str_`?

synthtool/gcp/common.py
```diff
@@ -153,10 +153,22 @@ def _load_partials(self, metadata: Dict):
...
+            next = ""
+            if i < len(str) - 1:
+                next = str[i + 1]
+            # "CoolACL" -> "Cool ACL" case.
+            if not re.match(r"[A-Z]", prev) or re.match(r"[a-z]", next):
+                str2 += " "
+            str2 += chr
         else:
             str2 += chr
+        prev = str2[len(str2) - 1]
```
tswast: I guess this is to handle when `prev` should be `" "`? I'd prefer if the above set a temporary variable for the string to append, then this becomes

```
output += to_append
prev = to_append
```

synthtool/__main__.py
tswast: Honestly that should be a linting error. It's not always guaranteed that submodules are actually loaded when you import a top-level module.

That said, I'd prefer that you instead change the above import to `import importlib.util` instead. `util` is too common of a name to be meaningful.

PR #226, opened 2019-04-18
feat: rework .nycrc to support --all

bcoe:
## what is this?

This reworks our `.nycrc` to better support the flag `--all`; this lets us know we've missed writing coverage for files:

Screen Shot 2019-04-18 at 3 47 18 PM

_Note: this update requires nyc@14._

## oddities

I think the rules:

```
"src/*{/*,/**/*}.js",
"src/*/v*/*.js"
```

were an attempt to not collect coverage in the `src/v1`, `src/v3beta1` directory, etc.

1. `"src/*{/*,/**/*}.js",` feels a little heavy handed, and was stopping us from collecting coverage for some libraries.
2. do we not want to collect coverage for the file3s in `src/v1`? seems like in a perfect world we'd eventually have tests that exercise these APIs?


synthtool/gcp/templates/node_library/.nycrc
```diff
@@ -1,28 +1,22 @@
 {
   "report-dir": "./.coverage",
-  "reporter": "lcov",
+  "reporter": ["text", "lcov"],
   "exclude": [
-    "src/*{/*,/**/*}.js",
-    "src/*/v*/*.js",
-    "test/**/*.js",
-    "build/test"
```
JustinBeckwith: I notice this doesn't do anything funky with TypeScript.  Are we getting coverage from what's in `build/src`, or from `src/`?  How do we know this isn't being counted twice?

bcoe: Coverage uses source-maps, so in reports it will point the user to the files in src/ as the files missing coverage; in practice it’s always the files in build/src that are executed, but will not show up in reports due to them including source-maps.

JustinBeckwith:
So this LGTM.  Just please submit a hammer PR first, then lets land this one.

PR #230, opened 2019-04-23
feat: add support for metadata comments in samples

bcoe:
This adds support for providing additional sample meta-information, based on comments embedded within sample files, like so:

```js
// sample-metadata:
//   title: ACL (Access Control)
//   description: Demonstrates setting access control rules.
```

Fields we currently intend to support:

* `title`: a human readable title for the sample.
* `description`: description of a sample.
* `usage`: usage instructions for a sample, e.g., `node sample.js --hello`.
* `api_version`: what version of the API does this sample apply to, e.g., v1beta, v2.

## Future Work

We would like to eventually also support a `snippets` field, which includes meta information about individual snippets, if multiple snippets are included in the sample file; this has not yet been implemented, as it's not currently used in Node's `samples/README.md` or `README.md`, which will be the first two consumers of this functionality.

## Note

It's our hope that in the future this meta information will be automatically populated as samples are generated.

CC: @beccasaurus who was also involved in this discussion.


tests/fixtures/samples/meta2.js
```diff
@@ -0,0 +1,29 @@
+/**
+ * Copyright 2017, Google, Inc.
```
busunkim96: Nit: Can we use the 2019 version of the license?

PR #232, opened 2019-04-25
Fix move excludes

igorbernstein2:
#48 introduced a bug where the exclude path (which is relative to a tracked path) is compared to a source path (which is relative to the source root).
This issue was unnoticed because it works ok when the source path has a single component like in firestore:
https://github.com/googleapis/nodejs-firestore/blob/89afae/synth.py#L19

But breaks when the source path has multiple components like in bigtable:
https://github.com/googleapis/google-cloud-java/blob/264643/google-cloud-clients/google-cloud-bigtable/synth.py#L30

And when the cwd doesn't match the tracked path

### DIFF
--- a/synthtool/transforms.py
+++ b/synthtool/transforms.py
@@ -116,8 +116,8 @@ def _copy_dir_to_existing_dir(
                 e
                 for e in excludes
                 if (
-                    Path(e).relative_to(".") == Path(rel_path)
-                    or Path(e).relative_to(".") == Path(rel_path) / name
+                    Path(e) == _tracked_paths.relativize(root)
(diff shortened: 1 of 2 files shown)


igorbernstein2:
This is currently blocking:
https://github.com/googleapis/google-cloud-java/pull/5034

igorbernstein2:
Thanks for reviewing! Does anything else need to happen for this to be picked up by the next client regen?

busunkim96:
@igorbernstein2 Autosynth runs synthtool from the master, so nothing else should be necessary. If it's okay to wait the changes should show up on the synth PRs tomorrow morning.  (If this needs to go out today we ca re-trigger the Node.js synth job.)

Sorry for the delay in reviewing!

PR #239, opened 2019-05-03
fix: move codecov token to node10

bcoe:
once we roll out Node 12, we should switch from 10 to 12.


bcoe:
@JustinBeckwith to simplify migration, let's provide the token to both Node 8 and Node 10 during transition, that way there's less of a race condition between this and the build script changes.

JustinBeckwith:
I'm totally cool with that. What I'm saying is that you want to use `repo` to submit the PR, so we have the PR title right, and guaranteed to be just the change you want.  Otherwise, you end with with 70 "PLEASE CHANGE ME" PRs :)

PR #240, opened 2019-05-04
fix: temporary fix for broken Windows image

bcoe:
this `test.bat` file will unblock our broken Windows builds, it's pretty hacky; let me explain:

1. the image we shipped has an ancient version nvm, which installs Node@10.15.3 successfully, but fails to install npm (this might be because npm was moved from npm/npm to npm/cli).
2. the 8.9.1 version of Node.js has npm working \o/ hooray; so we can just use this to install the latest npm in 10.15.3 -- wrong!
3. the version of npm that installs with 8.9.1 apparently can't be run by Node 10 ... go figure.

tldr; the fix uses v8.9.1 to install the latest npm, then switches to 10.15.3 and uses this npm to install the latest version of npm on 10.15.3 (I think the second install is the easiest way to get all the linking correct).

### DIFF
--- a/synthtool/gcp/templates/node_library/.kokoro/test.bat
+++ b/synthtool/gcp/templates/node_library/.kokoro/test.bat
@@ -12,16 +12,7 @@
 @rem See the License for the specific language governing permissions and
 @rem limitations under the License.
 
-@echo "Starting Windows build"
-
-cd /d %~dp0
+cd /d %~dp0	
 cd ..
 
-call npm install -g npm@latest || goto :error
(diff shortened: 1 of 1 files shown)


bcoe:
@jkwlui first thing next week I think we should pair and work on getting a better Windows image built.

This hack will unblock @alexander-fenster from not being able to run CI/CD on Node@8, and will unblock any other builds that are failing due to our bad Windows image.

alexander-fenster:
It worked!

Issue #130, opened 2018-11-14
Force synthtool to be run as a module.


theacodes:
Currently it's possible to run both `python3 synth.py` and `python3 -m synthtool`/`synthtool`. We should eliminate one of these, and I vote that we keep the latter.

SurferJeffAtGoogle:
The latter is the preferred way.  But people manually run synth.py infrequently now; I can live with the current state.

PR #243, opened 2019-05-14
Document passing a synthfile

beccasaurus:
TIL that you can explicitly pass a synthfile because someone asked if you could.

They searched the source code to find the answer.

It's documented via `synthtool --help` but they checked the README.

### DIFF
--- a/README.md
+++ b/README.md
@@ -44,6 +44,11 @@ Run `synthtool`:
 python3 -m synthtool
 ```
 
+> You can specify a specific synthfile via:
+> ```
+> python3 -m synthtool path/to/synth.py
+> ```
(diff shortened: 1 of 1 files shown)


busunkim96:
I think we eventually want to deprecate running synthtool this way (see #130), but this is fine for now. Thank you for all the doc updates!

beccasaurus:
Feedback from user:

> "it does not quite work.  synthtool puts the generated stuff in the directory where it is run
or maybe it's synth.py that's doing that"

How it works didn't meet their expectations.

Closing in favor of not documenting this (also because of #130) – before you merge 😛

PR #248, opened 2019-05-23
Python: Update noxfile to blacken samples directory.

busunkim96:


synthtool/gcp/templates/python_library/noxfile.py.j2
```diff
@@ -23,6 +23,11 @@ import nox
 
 LOCAL_DEPS = (os.path.join("..", "api_core"), os.path.join("..", "core"))
 
+BLACK_PATHS = ["docs", "google", "tests", "noxfile.py", "setup.py"]
+
+if os.path.exists("samples"):
```
beccasaurus: Alternatively...

You could add "samples" to BLACK_PATHS

and reject any item that doesn't exist (do an os.path.exists on all items in the array)

beccasaurus: Rather than treating "samples" as a special item.

I mean... what if "docs" doesn't exist!

It should. But. Just a thought!

busunkim96: The other items *have* to be around for the library to be valid, so I think I will leave it this way for now. Thank you for the suggestion!

PR #252, opened 2019-05-25
[feat] Add microgenerator support to synthtool.

lukesneeringer:
This commit adds microgenerator support to synthtool, and defines shortcut methods for the three languages where the ACT team currently publishes Docker images.

A couple of things were not entirely clear to me when I wrote this, so I am looking for some guidance:

* What is the appropriate testing regimen? I did not see any existing unit tests that seemed to exercise the `GAPICGenerator` or `Artman` classes, so I did not attempt to write any for mine either.
* It is not clear to me how you want to handle output directories. This might be because artman largely took the option out of your hands? It seems to me that the right thing to do is to make a temporary directory, generate code there, and copy out of it, but currently to do that correctly, it probably requires that the individual `synth.py` files handle the temporary directory context manager logic.


synthtool/gcp/gapic_microgenerator.py
```diff
@@ -0,0 +1,199 @@
...
+            )
+        if not tuple((googleapis / proto_path).glob("*.proto")):
+            raise FileNotFoundError(
+                f"Directory {(googleapis / proto_path)} exists, but no protos found."
+            )
+
+        # Ensure the desired output directory exists.
+        # If none was provided, create a temporary directory.
+        if not output_dir:
+            output_dir = tempfile.mkdtemp()
```
lukesneeringer: @crwilcox I think it is pretty likely that this is the wrong thing to do, but it is not obvious to me what a _better_ option is.

Here are a couple potential possibilities:

  * Return a `TemporaryDirectory` object _instead of_ a `Path`.
    * _Pros:_ Easy for `synth.py` files; they just use the context manager interface to copy out of the temporary directory and cleanup magically happens.
    * _Cons:_ Violates / changes the interface observed by `GAPICGenerator`.
  * Require an output directory.
    * _Pros:_ Unambiguous. If you create a temporary directory to handle this, you clean it up.
    * _Cons:_ Slightly more code than the option above (~1-2 more lines in each `synth.py` file). Slightly violates the interface by requiring `output_dir`.
  * What is here now.
    * _Pros:_ Holds to the existing interface.
    * _Cons:_ Very easy to leave temporary directories lying around.
  * Try to return a frankenstein object that subclasses both `Path` and `TemporaryDirectory`. (In this case, I would remove `output_dir` entirely and _only_ write to temp directories.)
    * _Pros:_ Holds to the existing interface.
    * _Cons:_ Seems dirty. Edge cases intuitively seem likely.

busunkim96: I am OK with the approach here. Synthtool already leaves behind generated libraries in `~/.cache/synthtool/googleapis/artman-genfiles/`.

In fact, we have a note in the README that recommends taking a look at that directory when you're doing a generation.

> SynthTool will run Artman which will create generated code that can be found at `~/.cache/synthtool/googleapis<-private>/artman_genfiles`. This is useful for figuring out what it is you need to copy for your specific library.

Would it be reasonable to have the microgenerator default to a similar location for predictability?

*For reference:*
Artman's output_dir is `root_dir / "artman-genfiles"`
https://github.com/googleapis/synthtool/blob/6f7311c3a6fc555f81fb0bcb08a4596e11e98608/synthtool/gcp/artman.py#L34-L43

`root_dir` is the path for `googleapis`.
https://github.com/googleapis/synthtool/blob/6f7311c3a6fc555f81fb0bcb08a4596e11e98608/synthtool/gcp/gapic_generator.py#L108-L112

And googleapis is cloned to a cache dir, by default.
https://github.com/googleapis/synthtool/blob/6f7311c3a6fc555f81fb0bcb08a4596e11e98608/synthtool/sources/git.py#L47-L55
https://github.com/googleapis/synthtool/blob/6f7311c3a6fc555f81fb0bcb08a4596e11e98608/synthtool/cache.py#L19

lukesneeringer: Thanks @busunkim96 for looking at this!

The new generators are protoc plugins, which require a specified directory to output to (that is the fundamental protoc plugin interface), so I can not have the generator itself default to any location.

However, I _can_ have synthtool pick a default location (e.g. `python-gapic-genfiles`) and behave this way if that is your preference. That seems odd to me, but you are the end user, and I defer to your preference. :)

Should I make that change? Happy to do so.

busunkim96: @lukesneeringer Yes please! I think it would be the least surprising for the microgenerator to also put files in the synthtool cache dir. Thank-you. :smile:

busunkim96:
If folks want the library to be in some other directory, that can be changed in a future PR.

PR #265, opened 2019-06-20
Add support for using local GAPIC generator

### LINKED ISSUE #250: Add option for local GAPIC generator (/toolkit)
Today, you can set `SYNTHTOOL_GOOGLEAPIS` and it'll use a local version of googleapis:

**[`gapic_generator.py`](https://github.com/googleapis/synthtool/blob/dd25cdf7cf775be96c488d42816ac31bef7db792/synthtool/gcp/gapic_generator.py#L27)**
```py
27: LOCAL_GOOGLEAPIS: Optional[str] = os.environ.get("SYNTHTOOL_GOOGLEAPIS")
```

### How does this work?

It adds a `--volume` to the docker invocation of the googleapis/artman Docker image.

The way you specify a local googleapis or GAPIC generator with the Docker artman image is by providing `-v` or `--volume`s which map these local directories on your PC to directories inside of the container.

If you provide your own path for `/googleapis`, then it will mount your local path to googleapis, else it will pull down the latest https://github.com/googleapis/googleapis repo and use that.

If you provide your own path for `/toolkit`, then it will mount your local path to gapic-generator, else it will pull down the latest https://github.com/googleapis/gapic-generator repo and use that.

### Feature request

Add the following to support a `SYNTHTOOL_GAPIC_GENERATOR` environment variable :)

----

**[`gapic_generator.py`](https://github.com/googleapis/synthtool/blob/dd25cdf7cf775be96c488d42816ac31bef7db792/synthtool/gcp/gapic_generator.py#L27)**
```py
28: LOCAL_GAPIC_GENERATOR: Optional[str] = os.environ.get("SYNTHTOOL_GAPIC_GENERATOR")
```

And pass it along as `-v [THIS PATH]:/toolkit` (see how we pass `/googleapis` as an example)

This is important so that we can run synthool to generate libraries using a LOCAL copy of the GAPIC generator :)

Sometimes we need to test using a version of GAPIC generator which has not yet been deployed in a new googleapis/artman DockerHub image.

----

#### Follow-up / Implementation Notes

 - Reading [`artman.py`](https://github.com/googleapis/synthtool/blob/dd25cdf7cf775be96c488d42816ac31bef7db792/synthtool/gcp/artman.py#L108) we don't actually map to `/googleapis`, we use `-w` to set our googleapis directory to be the working directory.
 - `-v f"{root_dir}:{root_dir}",`
 - `"-w", root_dir`
 - For a custom GAPIC generator, you would want Artman to have a property for the path to gapic_generator, if locally present, and (if locally present) pass:
 - `-v f"{local_gapic_generator}:/toolkit"`

As an example, here is the 'Usage' for how to invoke googleapis/artman with options:

#### Usage [googleapis/artman]

1. Pull down latest artman Docker image:
 - `docker pull googleapis/artman`
2. Build a library using its artman config yaml and the latest HEAD `googleapis/gapic-generator` (output to current directory)
 - ```
    docker run --rm                                     \
        --workdir /googleapis                           \
        --volume `pwd`:/googleapis/artman-genfiles      \
        --env RUNNING_IN_ARTMAN_DOCKER=True             \
        googleapis/artman /bin/bash -c                  \
        "artman --local --config /googleapis/google/cloud/kms/artman_cloudkms.yaml generate python_gapic"
    ```
3. If you’d like to use your own local version of googleapis or gapic-generator, add the following flags to the Docker command:
 - `--volume LOCAL_PATH_TO_GOOGLEAPIS:/googleapis`
 - `--volume LOCAL_PATH_TO_GAPIC_GENERATOR:/toolkit`
 - `--volume OUTPUT_DIR:/googleapis/artman-genfiles` (override where the output is written)

beccasaurus:
```
export SYNTHTOOL_GENERATOR=/path/to/gapic-generator
```

Fixes #250 

**See update to README below for description :)**

---

Refresher on how this is used:

 - I work on and with the monolithic generator.
 - Today, SynthTool runs artman which includes the latest version of gapic-generator (in the image)
 - Today, the easiest way for me to run SynthTool with changes to gapic-generator is to push a new version of artman, without much local testing
 - With this, I/we can run SynthTool locally while developing gapic-generator and verify functionality much easier! Very helpful :)

Considerations:

 - While today there is one "generator", that will not always be the case.
 - But the current codebase is authored based around artman which is based around the single generator use-case.
 - As such, I am not doing anything special. I've named the variable `SYNTHTOOL_GENERATOR` knowing full well that there will be a need to support N generators in the future :)

---

/cc @googleapis/samplegen

### DIFF
--- a/README.md
+++ b/README.md
@@ -247,6 +247,21 @@ Specify the path to`googleapis` in the environment variable `SYNTHTOOL_GOOGLEAPI
 export SYNTHTOOL_GOOGLEAPIS=path/to/local/googleapis
 ```
 
+### Local GAPIC Generator
+SynthTool supports generation from a local copy of [gapic-generator](https://github.com/googleapis/gapic-generator).
+Specify the path to`gapic-generator` in the environment variable `SYNTHTOOL_GENERATOR`.
+
+```
+export SYNTHTOOL_GENERATOR=path/to/local/gapic-generator
+```
+
+Don't forget to compile `gapic-generator` before running SynthTool.
+
+```
+cd path/to/local/gapic-generator
+./gradlew fatJar
+```
+
 ### Include .proto files
 SynthTool supports copying .proto API definition files from googleapis.
 
--- a/synthtool/gcp/artman.py
+++ b/synthtool/gcp/artman.py
@@ -58,7 +58,14 @@ class Artman:
     def docker_image(self) -> str:
         return self._docker_image_info()["RepoDigests"][0]
 
-    def run(self, image, root_dir, config, *args, generator_args=None):
+    def run(
+        self,
+        image,
+        root_dir,
+        config,
+        *args,
+        generator_dir=None,
+        generator_args=None):
         """Executes artman command in the artman container.
 
         Args:
@@ -72,6 +79,9 @@ class Artman:
             *args:
                 Arguments to artman that follow ``generate``. Defines which
                 artifacts to generate.
+            generator_dir (Optional[str]):
+                Path to local gapic-generator directory to use for generation.
+                By default, the latest version of gapic-generator will be used.
             generator_args (Optional[List[str]]):
                 Additional arguments to pass to the gapic generator, such as
                 ``--dev_samples``.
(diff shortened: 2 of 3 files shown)


busunkim96:
@beccasaurus This looks great! Would you mind doing a `nox -s blacken` to appease our lint checker? :grin:

beccasaurus:
@busunkim96 Done :) All green!

PR #267, opened 2019-06-24
Allow replacements to be marked required.

### LINKED ISSUE #266: Add ability to mark a replacement as required
For some libraries (like Grafeas), we are doing source replacement that must be done. It's possible that the source code changes (generator changes, formatting, etc) and the replacement fails to match. We already warn if the replacement doesn't happen which is great when it's for a bug fix. When it's a required replacement, we should give the option to make synth fail.

Perhaps, something like:
```python
s.replace(source_path, search, replace, required=True)
```

busunkim96:
Closes #266.


synthtool/transforms.py
```diff
@@ -213,7 +217,12 @@ def replace(
             log.info(f"Replaced {before!r} in {path}.")
 
     if not any_replaced:
-        log.warning(
-            f"No replacements made in {sources} for pattern {before}, maybe "
-            "replacement is not longer needed?"
-        )
+        if required:
+            log.error(
```
busunkim96: @theacodes Will this make the job 'failed' from Autosynth's perspective?

theacodes: I don't think so.

PR #268, opened 2019-06-25
feat: add anchor so that we can link directly to reference docs

bcoe:
Adds an anchor, so that when linking out to `googleapis.dev`, we can link directly to the section of the page that has reference docs.


bcoe:
@JustinBeckwith, I've got a mass PR ready to go, which uses `synthtool`.

PR #287, opened 2019-09-11
Java templates: update renovate config

chingor13:


synthtool/gcp/templates/java_library/renovate.json
```diff
@@ -24,6 +25,7 @@
     {
       "packagePatterns": ["^io.opencensus:opencensus-"],
       "groupName": "OpenCensus packages"
-    }
-  ]
-}
\ No newline at end of file
+    },
```
busunkim96: Is renovate OK with trailing commas?

chingor13: good catch thanks!

PR #292, opened 2019-09-20
Java: renovate config - use docker version scheme for guava

chingor13:
This forces renovate to keep the same version suffix. i.e. 28.0-android -> 28.1-android (rather that switching to -jre)

see https://github.com/renovatebot/config-help/issues/232#issuecomment-491826549


chingor13:
Closed in favor of #297

PR #293, opened 2019-09-21
fix: do not eslint in protos/

alexander-fenster:
We've got some generated stuff in `protos/`, such as `protos.js` and `protos.d.ts`, they are huge and we don't want `eslint` to check/fix them.

### DIFF
--- a/synthtool/gcp/templates/node_library/.eslintignore
+++ b/synthtool/gcp/templates/node_library/.eslintignore
@@ -2,3 +2,4 @@
 src/**/doc/*
 build/
 docs/
+protos/


alexander-fenster:
@bcoe @JustinBeckwith Can I please get your ✅ on this?

alexander-fenster:
@bcoe @JustinBeckwith ping :)

alexander-fenster:
I actually added `protos/` to the file manually for many (but not all) repos, I'll make sure I go one by one and see that no `CHANGE ME` gets in.

PR #306, opened 2019-10-14
feat: add helper for the most common Java library synth generation

chingor13:
This will greatly simplify our synth.py scripts across our split java repos. We do all of these steps when we generate a java versioned library.

### DIFF
--- a/synthtool/languages/java.py
+++ b/synthtool/languages/java.py
@@ -109,7 +109,7 @@ def fix_grpc_headers(grpc_root: Path, package_name: str) -> None:
 def gapic_library(
     service: str,
     version: str,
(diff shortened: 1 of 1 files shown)


synthtool/languages/java.py
```diff
@@ -94,6 +95,7 @@ def fix_proto_headers(proto_root: Path) -> None:
         f"{GOOD_LICENSE}{PROTOBUF_HEADER}",
     )
     s.replace([proto_root / "src/**/*Name.java"], BAD_LICENSE, GOOD_LICENSE)
```
busunkim96: You can pass a list to replace
```suggestion
     s.replace([proto_root / "src/**/*Name.java", proto_root / "src/**/*Names.java"], BAD_LICENSE, GOOD_LICENSE)
```

PR #308, opened 2019-10-17
fix: in Node.js repo-name and package-name can be different

bcoe:
This will address a bug with `samples/package.json` not being updated.

@busunkim96 this might be the case with Python as well? You might want to populate `packageName` in your config.


bcoe:
I'm going to make an effort to populate this automatically.

PR #313, opened 2019-10-18
feat(java): enable samplegen

chingor13:
Java has a special, deep directory structure for generated samples. Therefore, we need to make the directory structure configurable for` _include_samples()`.

### DIFF
--- a/synthtool/gcp/gapic_generator.py
+++ b/synthtool/gcp/gapic_generator.py
@@ -170,8 +170,20 @@ class GAPICGenerator:
             log.success(f"Placed proto files into {proto_output_path}.")
 
         if include_samples:
+            samples_root_dir = None
+            samples_resource_dir = None
+            if language == "java":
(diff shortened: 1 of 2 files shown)


busunkim96:
I think the mypy error might be coming from `f` being used twice as a name for the file handler

```py
            with open(googleapis_resources_yaml, "r") as f:
                resources_data = yaml.load(f, Loader=yaml.SafeLoader)
            resource_list = resources_data.get("sample_resources")
            for resource in resource_list:
                uri = resource.get("uri")
                if uri.startswith("gs://"):
                    uri = uri.replace("gs://", "https://storage.googleapis.com/")
                response = requests.get(uri, allow_redirects=True)
                download_path = samples_resources_dir / os.path.basename(uri)
                os.makedirs(samples_resources_dir, exist_ok=True)
                log.debug(f"Download {uri} to {download_path}")
                with open(download_path, "wb") as f:
                    f.write(response.content)
```
Could you try changing one of the `f`s to something else?

```py
                with open(download_path, "wb") as fd:
                    fd.write(response.content)
```

chingor13:
cc @beccasaurus

PR #317, opened 2019-10-25
feat(coverage): start providing coverage for src

bcoe:
excluding ` "**/src/**/v*/**/*.js"` prevents us from getting interesting insights into how our libraries are tested. As an example, I just ran `@google-cloud/translate` with this stanza removed, and it told me we don't call `getProjectId` in our unit tests:

example

I think we should cover `src` by default, and if there's a compelling reason to remove it for some projects we can exclude `.nycrc` in `synth.py`.


bcoe:
I'm going to hold off and land this after a mass update to our repos with automation.

bcoe:
we currently have per-repo updates on the wire, so I'm going to go ahead and land this.

PR #321, opened 2019-11-06
feat: pass extra protos to a microgenerator

alexander-fenster:
Sometimes we'll need to pass some extra proto files, such as `google/cloud/common_resources.proto`, to microgenerators. The idea is to mount all the extra files into the Docker image (e.g. `/extra/google/cloud/common_resources.proto`), and pass all of them as options (passing `-I /extra` for `protoc`, and then all the filenames).


bcoe:
@alexander-fenster looks like you will need to fix a couple linting issues.

PR #331, opened 2019-11-19
improvement(java_template): update mvn commands to skip enforcer checking for clirr and integration builds

BenWhitehead:


BenWhitehead:
@chingor13

PR #332, opened 2019-11-20
feat: refactor template metadata into classes

chingor13:
This allows languages to customize the loading of template metadata via classes/inheritance rather than messing around with dictionaries which are conditionally mucked with in private methods.

### DIFF
--- a/synthtool/gcp/common.py
+++ b/synthtool/gcp/common.py
@@ -12,20 +12,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-import json
-import os
-import re
-import yaml
 from pathlib import Path
-from typing import List, Dict
+from typing import List
 
 from synthtool.languages import node
 from synthtool.sources import templates
-from synthtool.gcp.templates.metadata import Metadata, JavaMetadata, NodejsMetadata
+from synthtool.gcp.templates.metadata import (
+    Metadata,
+    JavaMetadata,
+    NodejsMetadata,
+)
 from synthtool import __main__
 from synthtool import _tracked_paths
 from synthtool import metadata
-
 
 _TEMPLATES_DIR = Path(__file__).parent / "templates"
 _RE_SAMPLE_COMMENT_START = r"\[START \w+_quickstart\w*]"
--- a/synthtool/gcp/common.py
+++ b/synthtool/gcp/common.py
@@ -21,6 +21,7 @@ from synthtool.gcp.templates.metadata import (
     Metadata,
     JavaMetadata,
     NodejsMetadata,
+    decamelize,
 )
 from synthtool import __main__
 from synthtool import _tracked_paths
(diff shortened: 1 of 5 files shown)


synthtool/gcp/common.py
```diff
@@ -38,8 +41,8 @@ def __init__(self):
 
     def _generic_library(self, directory: str, **kwargs) -> Path:
         # load common repo meta information (metadata that's not language specific).
+
         if "metadata" in kwargs:
```
SurferJeffAtGoogle: Why do we search for "metadata" in kwargs?
Why not declare an argument metadata=None?

synthtool/gcp/common.py
```diff
@@ -12,24 +12,27 @@
...
 from synthtool import __main__
 from synthtool import _tracked_paths
 from synthtool import metadata
 
-
 _TEMPLATES_DIR = Path(__file__).parent / "templates"
 _RE_SAMPLE_COMMENT_START = r"\[START \w+_quickstart\w*]"
 _RE_SAMPLE_COMMENT_END = r"\[END \w+_quickstart\w*]"
 
+decamelize = decamelize
```
SurferJeffAtGoogle: What does this do?

synthtool/gcp/templates/metadata.py
```diff
@@ -0,0 +1,185 @@
...
+
+import requests
+
+_RE_SAMPLE_COMMENT_START = r"\[START \w+_quickstart\w*]"
+_RE_SAMPLE_COMMENT_END = r"\[END \w+_quickstart\w*]"
+
+
+class Metadata:
+    def __init__(self, initial_data: dict = {}):
+        self._data = initial_data if initial_data else {}
```
SurferJeffAtGoogle: I like how this code has been moved into a separate .py file, but If self._data is the only private data, then I actually prefer the functional structure of the original code.

synthtool/gcp/templates/metadata.py
```diff
@@ -0,0 +1,185 @@
...
+        self._load_from_files()
```
SurferJeffAtGoogle: This constructor calls a huge chunk of code.  Constructors are normally limited to initialization.  I don't think the constructor is the right place to invoke _load_from_files(), and it supports my claim above that using a class to represent this data is a mismatch.

synthtool/gcp/templates/metadata.py
```diff
@@ -0,0 +1,185 @@
...
+        with open(partials_file) as f:
+            self._data["partials"] = yaml.load(f, Loader=yaml.SafeLoader)
+
+    def __getitem__(self, key: str):
+        return self._data[key]
+
+
+class JavaMetadata(Metadata):
+    def __init__(self, initial_data: dict = {}):
+        Metadata.__init__(self, initial_data)
```
SurferJeffAtGoogle: Reusing code through inheritance is a known anti-pattern,
https://en.wikipedia.org/wiki/Composition_over_inheritance
further hinting that classes are a mismatch for this code.

synthtool/gcp/templates/metadata.py
SurferJeffAtGoogle: Here's a link that describes the issue:
https://dzone.com/articles/fixing-the-constructor-anti-pattern

Issue #216, opened 2019-03-28
synthtool keeps files that were removed in updated protos


jkwlui:
Encountered this [issue](https://github.com/googleapis/nodejs-talent/pull/29#issuecomment-477755487) generating the nodejs-talent library today (caught indirectly by Node.js's docs-test).

In the latest upstream commit to googleapis (https://github.com/googleapis/googleapis/commit/5fe8a4f65b8606b827cb28cd437cf4d42b4f9920), `ResumeService` was removed entirely from the proto definitions. I know at least in Node.js and Python, the files associated with `ResumeService` still lingers in the repo.

I had to remove the gapic and protos directories and re-run synthtool locally to me to see that the files are being deleted.

This should not be a breaking behaviour, because the removed service client is removed from the exports in `{version}/index.js`, so users would not be able to `require` the client anyways, unless as @busunkim96 points out if they do weird, unsupported things like requiring the client using the direct file system path.

We're currently adding a bunch of files in addition to those generated by synthtool, so a `rm -rf * && synthtool` is out of the question.

Proposals:
1. Record files generated in `synthtool.metadata` and if a new generation doesn't yield files previously generated, then remove them.

1. Specifically target generated gapic directories, knowing that we do not add additional files to them and generate into the directory from a clean state.

SurferJeffAtGoogle:
I implemented a solution by examining generated file time stamps but that solution was flawed because copying files usually preserves the time stamps.  So I rolled back the solution.

@fhinkel and @bcoe and @JustinBeckwith recently suffered due to this issue still existing, so I'm re-opening it and trying a new strategy: using https://pypi.org/project/watchdog/ to observe file system events.

PR #342, opened 2019-12-09
feat: remove obsolete files during synthesis

SurferJeffAtGoogle:
Fixes https://github.com/googleapis/synthtool/issues/216

The new code implements a very conservative strategy when deciding which files were obsolete.

A file is deemed obsolete only if:
1. It was generated in the previous run of synthtool.
2. It was not touched after the current run of synthtool.

Implementing this strategy requires keeping a list of all files that were generated in the metadata.


tests/test_metadata.py
```diff
@@ -91,3 +93,49 @@ def test_write(tmpdir):
...
+    with open(b_path, "wt") as file:
+        file.write("b")
+    time.sleep(1)
+    after_b_before_c = time.time()
+    time.sleep(1)
+    c_path = os.path.join(codedir, "c")
+    with open(c_path, "wt") as file:
+        file.write("c")
+
+    # Confirm add_new_files found the new files and ignored the old one.
```
bcoe: I want be tempted to break this out into a couple assertions, perhaps you could create a `create_test_directory` helper, and then have a:

`test_add_new_files`, and `remove_obsolete_files` assertion?

SurferJeffAtGoogle: Done.

tests/test_metadata.py
```diff
@@ -91,3 +95,76 @@ def test_write(tmpdir):
...
+    metadata.add_new_files(
+        source_tree_fixture.after_a_before_b, source_tree_fixture.srcdir
+    )
+    assert 2 == len(metadata._metadata.new_files)
+    new_file_paths = [new_file.path for new_file in metadata._metadata.new_files]
+    assert os.path.relpath(source_tree_fixture.b_path) in new_file_paths
+    assert os.path.relpath(source_tree_fixture.c_path) in new_file_paths
+
+
+def test_old_file_removed(source_tree_fixture):
```
bcoe: neat, cool to see how fixtures work.

PR #343, opened 2019-12-13
fix(docs): include sample dependencies

pheonixblade9:
Update docs to include sample dependencies to clarify sample run instructions


SurferJeffAtGoogle:
Why is this PR labeled with blocked?

JustinBeckwith:
If I had to guess.... 

We usually like to make these changes locally using [repo](https://github.com/googleapis/github-repo-automation/), and then send out a blanket PR for all 78 repositories at once.  If we let synthtool do it, it doesn't preserve the commit message, so we have to edit 78 PRs and hope that there are no other related changes in the morning. 

But only @bcoe really knows 😆

PR #345, opened 2019-12-13
feat: remove obsolete files during synthesis again

SurferJeffAtGoogle:


bcoe:
@SurferJeffAtGoogle the new test case looks great, but I'm not seeing where we handle the bad symlinks (I see we handle missing files in a `try`/`catch`). Was I wrong about the cause of of the failure?

SurferJeffAtGoogle:
> @SurferJeffAtGoogle the new test case looks great, but I'm not seeing where we handle the bad symlinks (I see we handle missing files in a `try`/`catch`). Was I wrong about the cause of of the failure?

When the file is a bad symlink, the following try & catch catches it:
```py
            try:
                mtime = os.path.getmtime(filepath)
            except FileNotFoundError:
                log.warning(
                    f"FileNotFoundError while getting modified time for {filepath}."
                )
```

The test creates a bad symlink, and FileNotFoundError is the exception that gets thrown.

SurferJeffAtGoogle:
> Awesome, `FileNotFoundError` isn't the error I would have guessed, but the test case looks really solid.

Thank you for the review.  Merging...

PR #348, opened 2019-12-17
feat: return # of replacements

### LINKED ISSUE #266: Add ability to mark a replacement as required

SurferJeffAtGoogle:
Supercedes #267
Fixes #266 with less code.  The caller can decide if replacing nothing is a fatal error and throw an exception.


synthtool/transforms.py
```diff
@@ -185,35 +185,37 @@ def _replace_in_file(path, expr, replacement):
...
+        if count:
+            fh.seek(0)
+            fh.write(content)
+            fh.truncate()
+        return count
 
 
 def replace(
     sources: ListOfPathsOrStrs, before: str, after: str, flags: int = re.MULTILINE
 ):
```
busunkim96:
```suggestion
) -> int:
```

PR #361, opened 2019-12-30
feat(java): add helper for bazel generation

chingor13:
Some of the new clients do not use the artman docker image and instead use bazel. For these libraries, we can share common behavior between the artman generation and the bazel generator for things like code formatting and fixing license headers.

This PR pulls the shared code of copying to the correct destination folder into a private `_common_generation()` function which is shared between the `java.gapic_library()` and `java.bazel_library()` functions.

Fixes: https://github.com/googleapis/java-secretmanager/issues/11

### DIFF
--- a/synthtool/languages/java.py
+++ b/synthtool/languages/java.py
@@ -108,17 +108,16 @@ def fix_grpc_headers(grpc_root: Path, package_name: str) -> None:
         f"{GOOD_LICENSE}package {package_name};",
     )
 
+
 def _common_generation(
-    service: str,
-    version: str,
-    library: Path,
-    package_pattern: str,
-    suffix: str = ''
-    ):
+    service: str, version: str, library: Path, package_pattern: str, suffix: str = ""
+):
 
     package_name = package_pattern.format(service=service, version=version)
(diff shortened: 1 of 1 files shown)


chingor13:
@SurferJeffAtGoogle ping on this one - this is needed for a few of our new proto annotation-only libraries

synthtool/languages/java.py
```diff
@@ -109,54 +109,41 @@ def fix_grpc_headers(grpc_root: Path, package_name: str) -> None:
...
-    version: str,
-    config_pattern: str = "/google/cloud/{service}/artman_{service}_{version}.yaml",
-    package_pattern: str = "com.google.cloud.{service}.{version}",
-    gapic: gcp.GAPICGenerator = None,
-    **kwargs,
-) -> Path:
-    if gapic is None:
-        gapic = gcp.GAPICGenerator()
+def _common_generation(
+    service: str, version: str, library: Path, package_pattern: str, suffix: str = ""
```
SurferJeffAtGoogle: Please add doc comments for the 3 functions touched in this PR.

PR #362, opened 2020-01-02
build: exclude .d.ts files from coverage

schmidt-sebastian:
These files are not executable, so their coverage will always be 0.

This increases coverage in Firestore by 8% (https://github.com/googleapis/nodejs-firestore/pull/852)


schmidt-sebastian:
@bcoe - Can you take a look?

PR #366, opened 2020-01-08
fix: ignore files ignored by git when tracking obsolete files

SurferJeffAtGoogle:


synthtool/metadata.py
```diff
@@ -100,26 +117,61 @@ def write(outfile: str = "synth.metadata") -> None:
...
+        if ".git" not in pathlib.Path(file_path).parts
+    ]
+    # Write the files to a temporary text file.
+    with tempfile.TemporaryFile("w+b") as f:
+        for file_path in nongit_file_paths:
+            f.write(_git_slashes(file_path).encode("utf-8"))
+            f.write("\n".encode("utf-8"))
+        # Invoke git.
+        f.seek(0)
+        git = shutil.which("git")
```
bcoe: I like this approach, you shell out to `git` rather than trying to parse the `.gitignore` file (as an alternative approach).

SurferJeffAtGoogle: Thanks.  Fewer lines of code and guaranteed to behave the same as git.

SurferJeffAtGoogle:
> This is looking reasonable to me, should we go back to setting `_track_obsolete_files` to `True` by default, now that we handle ignored files?

First, let's manually turn it on for the few repos where we observed problems.  Once we're confident it fixed the issues, let's flip the default to on.

PR #370, opened 2020-01-13
feat: record git log in metadata for each git source

SurferJeffAtGoogle:


synthtool/gcp/gapic_bazel.py
```diff
@@ -200,7 +200,7 @@ def _clone_googleapis(self):
 
         else:
             log.debug("Cloning googleapis.")
-            self._googleapis = git.clone(GOOGLEAPIS_URL, depth=1)
+            self._googleapis = git.clone(GOOGLEAPIS_URL)
```
chingor13: Note that the entire git history may be huge. Maybe it doesn't matter that much as the cache directory is shared across the entire synth run.

bcoe:
@chingor13 I had this on my mind too, the git history being giant, but if we were to do something like `depth=50`, does it preserve the history appropriately?

synthtool/gcp/gapic_bazel.py
SurferJeffAtGoogle: In the case where the git history is huge, what is the negative effect?  Will generation fail?

SurferJeffAtGoogle: Also, notice in git.py, I only download the single branch master, so it won't download the entire history of the repo.

chingor13: It's probably premature optimization and won't really matter in practice. If it is an issue, we can revisit.

synthtool/gcp/gapic_generator.py
```diff
@@ -229,7 +229,7 @@ def _clone_googleapis_private(self):
 
         else:
             log.debug("Cloning googleapis-private.")
-            self._googleapis_private = git.clone(GOOGLEAPIS_PRIVATE_URL, depth=1)
```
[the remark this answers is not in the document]
SurferJeffAtGoogle: Yes.

SurferJeffAtGoogle:
Thank you for the reviews and the approval.  I don't want to merge now and accidentally break library generation over a three day weekend, so I'll merge this on Tuesday.

PR #372, opened 2020-01-14
feat(java): add helper for fetching the latest version of a Maven artifact

chingor13:
As we move to templatize our java README files, we will need the ability to fetch the latest version of Maven artifacts from Maven Central.

This PR adds the ability to fetch and parse the latest version from Maven Central.

### DIFF
--- a/synthtool/languages/java.py
+++ b/synthtool/languages/java.py
@@ -130,7 +130,7 @@ def latest_maven_version(group_id: str, artifact_id: str) -> Optional[str]:
     )
     response = requests.get(url)
     response.raise_for_status()
-    return version_from_maven_metadata(response.content.decode("utf-8"))
+    return version_from_maven_metadata(response.content.decode(response.encoding))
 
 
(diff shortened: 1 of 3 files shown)


synthtool/languages/java.py
```diff
@@ -109,6 +111,46 @@ def fix_grpc_headers(grpc_root: Path, package_name: str) -> None:
...
+    Returns:
+        The latest version of the artifact as a string or None
+    """
+    group_path = "/".join(group_id.split("."))
+    url = (
+        f"https://repo1.maven.org/maven2/{group_path}/{artifact_id}/maven-metadata.xml"
+    )
+    response = requests.get(url)
+    response.raise_for_status()
+    return version_from_maven_metadata(response.content.decode("utf-8"))
```
SurferJeffAtGoogle: Are you sure it will be utf-8?  Can we check the content-encoding response header?

chingor13: Done. Using `response.text` which uses requests' best guess at the content encoding by parsing the response headers.

PR #382, opened 2020-01-21
feat: report combined commit log for all repos in metadata

SurferJeffAtGoogle:
This change takes the git history of all the repos, and packs it into a single string in synth.metadata.

Later, autosynth will pull the string out of synth.metadata, and add it to the PR description.


bcoe:
This is a complex enough change that I think I will need to read it through with fresh eyes tomorrow. Could I bother you to update the description with some details about the feature? Is it adding additional context to the PR itself, or to the `synth.metadata`?

I think if I'm following correctly, this will allow us to give better contextual information to folks in a mono repo?

SurferJeffAtGoogle:
> This is a complex enough change that I think I will need to read it through with fresh eyes tomorrow. Could I bother you to update the description with some details about the feature? Is it adding additional context to the PR itself, or to the `synth.metadata`?
I updated the PR description.

> I think if I'm following correctly, this will allow us to give better contextual information to folks in a mono repo?
Both mono repos and split repos.

synthtool/metadata.py
```diff
@@ -202,10 +214,27 @@ def __exit__(self, type, value, traceback):
...
 
+def _get_commit_log_since(path, sha):
+    """Invokes git to get the commit log in the given path since the given sha.
+
+    Returns:
+        string, the git log.
+    """
+    git = shutil.which("git")
+    output = subprocess.run(
+        [git, "-C", path, "log", "--pretty=%H%n%B", "--no-decorate", f"{sha}..HEAD"],
```
bcoe: For `googleapis/googleapis` this ends up being a fairly large list of updates, because it will include every update to the repo, what if we were clever and filtered the PRs to a sub-directory in `googleapis`, based on the API name?

I think we could potentially piggyback off the variables being passed in to the generator like `proto_path`.

synthtool/metadata.py
```diff
@@ -248,19 +263,60 @@ def _get_git_source_map(metadata) -> Dict[str, object]:
...
+    Combine the commit logs of all the git sources into one commit log.
+    """
+    text = []
+    for source in _metadata.sources:
+        if source.git and source.git.log:
+            git_source = source.git
+            name = git_source.name
+            text.append(f"Changes in {name}:")
+            text.append(git_source.log)
+    _metadata.combined_commit_log = "\n".join(text)
```
bcoe: I was thinking it might make sense to populate this as an array rather than a `\n` separated string, might make it easier if we want to manipulate the data.

synthtool/metadata.py
bcoe: alternatively, if piggybacking is a pain, perhaps we could populate an additional variable in `synth.py`.

SurferJeffAtGoogle:
There is a way to 100% accurately filter the upstream changes that caused changes in the current repo.  It's way more complicated, but if this PR is going to cause too much noise, then I think I need to hunker down and go for the bigger solution.

bcoe:
@SurferJeffAtGoogle up to you, I think it will provide more value if we can filter to the specific APIs, but if we want to get there incrementally and try this out that's fine by me too :smile: 

I just noticed that there was quite a bit of data that ended up in `synth.metadata` when I was doing my testing.

Issue #390, opened 2020-01-30
Generation for many node libraries is broken


JustinBeckwith:
I've seen this across a variety of repos, and I think I'm ready to call this a bug.  Running `synthtool` is deleting a bunch of files that shouldn't be deleted.  

To reproduce:
- clone https://github.com/googleapis/nodejs-datalabeling
- install the latest synthtool
- run synthtool

The net result is all of these files getting deleted.

```

Changes not staged for commit:
  (use "git add/rm <file>..." to update what will be committed)
  (use "git restore <file>..." to discard changes in working directory)
	deleted:    .github/CONTRIBUTING.md
	deleted:    .github/ISSUE_TEMPLATE.md
	deleted:    .gitignore
	deleted:    .kokoro/.gitattributes
	deleted:    .kokoro/continuous/node8/docs.cfg
	deleted:    .kokoro/continuous/node8/lint.cfg
	deleted:    .kokoro/continuous/node8/samples-test.cfg
	deleted:    .kokoro/continuous/node8/system-test-grpcjs.cfg
	deleted:    .kokoro/continuous/node8/system-test.cfg
	deleted:    .kokoro/presubmit/node8/docs.cfg
	deleted:    .kokoro/presubmit/node8/lint.cfg
	deleted:    .kokoro/presubmit/node8/samples-test.cfg
	deleted:    .kokoro/presubmit/node8/system-test-grpcjs.cfg
	deleted:    .kokoro/presubmit/node8/system-test.cfg
	deleted:    .kokoro/release/common.cfg
	deleted:    .repo-metadata.json
	deleted:    CHANGELOG.md
	deleted:    package.json
	modified:   protos/protos.d.ts
	modified:   protos/protos.js
	modified:   protos/protos.json
	deleted:    samples/.eslintrc.yml
	deleted:    samples/package.json
	deleted:    samples/quickstart.js
	deleted:    samples/test/test.js
	deleted:    src/index.ts
	modified:   src/v1beta1/data_labeling_service_proto_list.json
	deleted:    src/v1beta1/doc/google/cloud/datalabeling/v1beta1/doc_annotation.js
	deleted:    src/v1beta1/doc/google/cloud/datalabeling/v1beta1/doc_annotation_spec_set.js
	deleted:    src/v1beta1/doc/google/cloud/datalabeling/v1beta1/doc_data_labeling_service.js
	deleted:    src/v1beta1/doc/google/cloud/datalabeling/v1beta1/doc_data_payloads.js
[...]

PR #391, opened 2020-01-30
fix: turn off obsolete file tracking by default

SurferJeffAtGoogle:
Fixes https://github.com/googleapis/synthtool/issues/390

The file system's timestamps are not reliable; there is no way synth.py should have ever ended up in the list of generated files, as happened here: https://github.com/googleapis/nodejs-datalabeling/blob/95d32ddc0b8389dc28ca47951da7019fe3451e0d/synth.metadata#L367

### DIFF
--- a/tests/test_metadata.py
+++ b/tests/test_metadata.py
@@ -263,7 +263,7 @@ def preserve_track_obsolete_file_flag():
 
 
 def test_track_obsolete_files_defaults_to_true(preserve_track_obsolete_file_flag):
-    assert metadata.should_track_obsolete_files()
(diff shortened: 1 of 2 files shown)


tests/test_metadata.py
```diff
@@ -263,7 +263,7 @@ def preserve_track_obsolete_file_flag():
 
 
 def test_track_obsolete_files_defaults_to_true(preserve_track_obsolete_file_flag):
```
chingor13: nit: should probably rename the test too

SurferJeffAtGoogle: Done.

PR #398, opened 2020-02-03
fix: exception occurring due to missing "repo" field

bcoe:
Some of our Node.js repos started failing due to "repo" no longer being defaulted to an empty object, see: https://github.com/googleapis/gaxios/issues/231.

Would it be alright if we continue populating an empty object for the time being?

### DIFF
--- a/synthtool/gcp/common.py
+++ b/synthtool/gcp/common.py
@@ -112,9 +112,7 @@ class CommonTemplates:
         # metadata, so we don't need to do it again or overwrite it. Also, only
         # set the "repo" key if data is available.
         if "repo" not in metadata:
-            repo_metadata = _load_repo_metadata()
-            if repo_metadata:
-                metadata["repo"] = repo_metadata
+            metadata["repo"] = _load_repo_metadata()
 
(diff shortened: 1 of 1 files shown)


synthtool/gcp/common.py
```diff
@@ -112,9 +112,7 @@ def _load_generic_metadata(self, metadata: Dict):
         # metadata, so we don't need to do it again or overwrite it. Also, only
         # set the "repo" key if data is available.
```
chingor13: This last sentence would no longer apply

synthtool/gcp/common.py
```diff
@@ -250,4 +248,4 @@ def _load_repo_metadata(metadata_file: str = "./.repo-metadata.json") -> Optiona
     if os.path.exists(metadata_file):
         with open(metadata_file) as f:
             return json.load(f)
-    return None
+    return {}
```
chingor13: If this only ever returns a dictionary, we should change the type hint from `Optional[Dict]` -> `Dict`.

bcoe:
@chingor13 fixed :smile:

PR #399, opened 2020-02-05
refactor: remove code that tracks obsolete files

SurferJeffAtGoogle:
In the wild, the file system's timestamps can't be trusted.  Perhaps there are unknown processes opening the files for write.  So the strategy of checking file timestamps to see what has changed won't work.

### DIFF
--- a/synthtool/metadata.py
+++ b/synthtool/metadata.py
@@ -13,6 +13,7 @@
 # limitations under the License.
 
 import datetime
+import deprecation
 import locale
 import os
 import pathlib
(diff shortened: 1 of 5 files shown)


synthtool/metadata.py
```diff
@@ -118,26 +117,7 @@ def write(outfile: str = "synth.metadata") -> None:
...
-    obsolete_files = old_files - new_files
-    for file_path in git_ignore(obsolete_files):
-        try:
-            log.info(f"Removing obsolete file {file_path}...")
-            os.unlink(file_path)
-        except FileNotFoundError:
-            pass  # Already deleted.  That's OK.
-
-
+@deprecation.deprecated(deprecated_in="2020.02.04")
```
bcoe: I like this `@deprecation` mix-in.

PR #401, opened 2020-02-07
feat: allow autosynth to pass in pre-cloned repos

SurferJeffAtGoogle:
via an environment variable


synthtool/sources/git.py
```diff
@@ -141,3 +149,29 @@ def extract_commit_message_metadata(message: str) -> Dict[str, str]:
...
+
+
+PRECLONE_MAP_HELP = """
+A preclone map file contains a list of git repos that have already been cloned.
+Its contents are json and look like this:
+{
+  "http://github.com/googleapis/nodejs-vision.git": "/tmp/nodejs-vison",
+  "http://github.com/googleapis/googleapis": "/tmp/googleapis"
+}
+"""
```
chingor13: Should this be at the top?

synthtool/sources/git.py
```diff
@@ -141,3 +149,29 @@ def extract_commit_message_metadata(message: str) -> Dict[str, str]:
...
 
     return metadata
+
+
+PRECLONE_MAP_HELP = """
+A preclone map file contains a list of git repos that have already been cloned.
+Its contents are json and look like this:
+{
+  "http://github.com/googleapis/nodejs-vision.git": "/tmp/nodejs-vison",
+  "http://github.com/googleapis/googleapis": "/tmp/googleapis"
```
chingor13: nit: should these examples use `https`?

synthtool/sources/git.py
```diff
@@ -141,3 +149,29 @@ def extract_commit_message_metadata(message: str) -> Dict[str, str]:
...
+    if not preclone_map_path:
+        return {}
+    with open(preclone_map_path, "r") as json_file:
+        return json.load(json_file)
+
+
+def get_preclone(url: str) -> Optional[str]:
+    """Finds a pre-cloned git repo in the preclone map."""
+    preclone_map = _get_preclone_map()
+    return preclone_map.get(url)
```
chingor13: Do we want to try and be lenient on the format of the git url? i.e. `git@github.com:owner/repo.git` vs `https://github.com/owner/repo` vs `https://github.com/owner/repo.git`

synthtool/sources/git.py
```diff
@@ -45,26 +49,30 @@ def make_repo_clone_url(repo: str) -> str:
 
 
 def clone(
-    url: str,
-    dest: pathlib.Path = None,
-    committish: str = "master",
-    force: bool = False,
+    url: str, dest: pathlib.Path = None, committish: str = None, force: bool = False,
```
chingor13: Can we backfill some documentation here. It looks like we're changing the behavior of the default value of `committish`. It's now ignored if there's an existing clone specified. Otherwise, it defaults to "master"

synthtool/__main__.py
```diff
@@ -65,14 +64,18 @@ def extra_args() -> List[str]:
...
-    """Synthesizes source code according to the instructions in synthfile arg.
+    f"""Synthesizes source code according to the instructions in synthfile arg.
 
     Optional environment variables:
       SYNTHTOOL_ARTMAN_VERSION:  The version of artman to use.
       SYNTHTOOL_GOOGLEAPIS:      Path to local clone of https://github.com/googleapis/googleapis
       SYNTHTOOL_GENERATOR:       Path to local gapic-generator directory to use for generation.
                 By default, the latest version of gapic-generator will be used.
       AUTOSYNTH_USE_SSH:         Access github repos via ssh instead of https.
+      {git.PRECLONE_MAP_ENVIRONMENT_VARIABLE}:  Path to a json file.
```
bcoe: this seems like a useful feature, are you thinking that the server that runs `autosynth` will be stateful, and have the repos cloned already? or is this more for development machines.

SurferJeffAtGoogle: This is purely for autosynth.

SurferJeffAtGoogle:
Thank you for all your review comments.  I'll accommodate them, and I also plan to rename `preclones.json` to `preconfig.json` and make its structure more general so that we can also specify things like generator versions.

synthtool/sources/git.py
SurferJeffAtGoogle: Moved to top.

synthtool/sources/git.py
SurferJeffAtGoogle: These examples no longer exist because it's documented preconfig.proto.

synthtool/sources/git.py
SurferJeffAtGoogle: The plan is that autosynth alone will supply a preconfig.json file.  It will populate the preconfig file by looking at the previous synthtool's output synthtool.metadata file.  Therefore the urls should match exactly.  Do you see another use case or an opportunity for mismatch?

synthtool/sources/git.py
SurferJeffAtGoogle: I added a complete docstring.

I don't see the path through the code where committish gets ignored.

SurferJeffAtGoogle:
Ready for another look.

synthtool/sources/git.py
chingor13: The default value for `committish` is only "master" for the non-preclone scenario. This is probably fine as if you are providing a pre-cloned repo, then presumably the current state of that directory acceptable

SurferJeffAtGoogle: Indeed, using the current state of that directory is required for the autosynth run-at-every-commit strategy to work.

tests/test_git.py
```diff
@@ -67,3 +75,52 @@ def test_extract_commit_message_metadata():
...
+        # Preserve the original environment variables.
+        self.env = copy.copy(os.environ)
+        return super().setUp()
+
+    def tearDown(self):
+        os.environ = self.env
+        return super().tearDown()
+
+    def testClone(self):
+        local_directory = git.clone("https://github.com/googleapis/nodejs-vision.git")
```
bcoe: by default what directory path will this be cloned into, curious if it also ends up in `tmp`.

SurferJeffAtGoogle: It gets cloned into `~/.cache/synthtool/nodejs-vision`

PR #402, opened 2020-02-07
feat: generalize snippet/sample parsing and use in java templates

chingor13:
Not quite ready for review yet:

- [ ] add more tests
- [ ] make sure nodejs templates still work with their custom logic


chingor13:
Going to break this into smaller separate PRs to make reviewing easier

PR #403, opened 2020-02-11
feat: generalize samples/snippet parsing into new modules

chingor13:
Adds 2 new modules:

1. `synthtool.gcp.samples` - logic for reading samples from source and parsing sample metadata
1. `synthtool.gcp.snippets` - logic for parsing snippets from source

Samples and snippets are used for template generation and cloud.google.com documentation. These helpers will be used by the template logic, but will be done separately to make the PR easier to review. 

Notes:
* We have not yet refactored the templating functions to use this generalized implementation. Thus there is duplication of logic which will be removed when we refactor the templates to use these modules.
* I moved the fixture files into a nested structure to give us more space to define test fixtures (rather than assuming all fixtures live within a node repo setup)

See https://github.com/chingor13/synthtool/pull/1 for part 2 - refactoring the nodejs templates and adding java template support.


synthtool/gcp/samples.py
```diff
@@ -0,0 +1,83 @@
...
+# limitations under the License.
+
+import glob
+import re
+import os
+import yaml
+from typing import List, Dict
+
+
+def _read_sample_metadata_comment(sample_file: str) -> Dict:
```
SurferJeffAtGoogle: The PR describes the new code as duplicating existing logic.  Where is the existing logic?

synthtool/gcp/snippets.py
```diff
@@ -0,0 +1,72 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import glob
+import re
+from typing import Dict, List
+
+OPEN_SNIPPET_REGEX = r".*\[START ([a-z0-9_]+)\].*$"
+CLOSE_SNIPPET_REGEX = r".*\[END ([a-z0-9_]+)\].*$"
```
SurferJeffAtGoogle: What about exclude regions like this one?
https://github.com/GoogleCloudPlatform/python-docs-samples/blob/6879d73123135459fa45a36211505d1ead37978a/appengine/standard/endpoints-frameworks-v2/echo/app.yaml#L7

Instead of re-implementing the logic from the doc tag parser, can we re-use the same code as the doc tag parser?

synthtool/gcp/samples.py
chingor13: It's currently in the [`common.CommonTemplates` class](https://github.com/googleapis/synthtool/blob/2e9bcac85a01d7d4fcc9fd0b06d80f75207e01f3/synthtool/gcp/common.py#L144-L168)

synthtool/gcp/snippets.py
chingor13: I didn't know that exclude tag was a thing, however it looks like a custom, per-widget regex supplied in the content template. We also haven't open-sourced the framework for cloud.google.com and that logic is included there.

synthtool/gcp/samples.py
chingor13: Note that the original is removed in the part 2 draft linked in the description.

PR #404, opened 2020-02-11
build: introduce GitHub actions for unit tests

bcoe:
this introduces GitHub actions for unit tests specifically (we're still using kokoro for nightly tests, sample tests, system tests).


feywind:
Is this the pre Jinja template update for escaping braces?

PR #405, opened 2020-02-12
refactor: use samples/snippets modules for nodejs templates

chingor13:
* Moves custom node template logic into the `synthtool.languages.node` module.
* Adds a test for the custom quickstart logic handling
* Removes duplicated sample/snippet logic from `synthtool.gcp.common.CommonTemplates`

Note that the difference in implementation for the sample parsing - the samples module returns the full file path from the repo root, rather than just the filename. Thus, the README template is adjusted to not assume the files are in the samples directory.


synthtool/gcp/common.py
```diff
@@ -114,79 +127,6 @@ def _load_generic_metadata(self, metadata: Dict):
...
-                    # only add quickstart file to samples list if code sample is found.
-                    if file == "quickstart.js" and not metadata.get("quickstart", None):
-                        continue
-                    sample_metadata = {"title": decamelize(file[:-3]), "file": file}
-                    sample_metadata.update(
-                        self._read_sample_metadata_comment(samples_dir, file)
-                    )
-                    metadata["samples"].append(sample_metadata)
-
-    def _read_sample_metadata_comment(self, samples_dir: Path, file: str) -> Dict:
```
bcoe: thanks for taking on this refactor :ok_hand:

tests/test_node.py
```diff
@@ -0,0 +1,52 @@
...
+    metadata = node._template_metadata()
+
+    # should not have populated the quickstart for the README
+    assert not metadata["quickstart"]
+
+    # should not have a link to the quickstart in the samples
+    sample_names = list(map(lambda sample: sample["file"], metadata["samples"]))
+    assert "samples/quickstart.js" not in sample_names
+
+    os.chdir(cwd)
```
chingor13: @bcoe Does this correctly assert the expected behavior of the quickstart sample/snippets?

bcoe: this seems reasonable to me, we remove the quickstart from the table of contents, but we show it as an example in the README.

synthtool/gcp/common.py
```diff
@@ -84,7 +82,7 @@ def node_library(self, **kwargs) -> Path:
             if "samples/README.md" not in self.excludes:
                 self.excludes.append("samples/README.md")
 
-        kwargs["metadata"] = node.read_metadata()
+        kwargs["metadata"] = node._template_metadata()
```
SurferJeffAtGoogle: In python, members named with a leading _ are intended to be private.  Why is this code calling a private method?

chingor13:
Note: I fixed the 2 errors above and added tests for those cases. I ran this branch against repo-automation-bots and nodejs-asset as sanity checks for no changes.

synthtool/gcp/common.py
chingor13: The idea of naming node._template_metadata() was for it to be private for users of synthtool, not necessarily for within the package. If that's incorrect, then we can make it public.

SurferJeffAtGoogle: Hmmm.  The fact that common.py is importing specific code from node.py is a red flag, but I realize that precedes this pull request.

I'd rother see `_template_metadata` renamed `template_metadata`.

chingor13: It's not a huge deal, I was just worried about the public surface area of these modules in case people are using them directly in their synth files. For this case, it could be reasonable for someone to need access to these variables anyways.

synthtool/languages/node.py
```diff
@@ -42,6 +44,39 @@ def read_metadata():
...
+        * quickstart (str): Contents of the quickstart snippet if available, otherwise, ""
+        * samples (List[Dict[str, str]]): List of available samples. See synthtool.gcp.samples.all_samples()
+    """
+    metadata = {}
+    try:
+        metadata = read_metadata()
+    except FileNotFoundError:
+        pass
+
+    all_samples = samples.all_samples(["samples/*.js"])
```
bcoe: I think we end up adding the prefix `samples` twice somewhere around here.

chingor13: Missed one in the `samples/README.md` template

PR #406, opened 2020-02-12
feat(java): add samples table, use install snippet if available in README

chingor13:
Uses the samples/snippets modules to provide samples to link for the README.


chingor13:
This one is actually blocked by #405. Without #405, the `samples` metadata is overwritten by the nodejs logic

PR #409, opened 2020-02-14
feat: treat templates as an external git source

chingor13:
You can provide a local checkout of the template directory via the
SYNTHTOOL_TEMPLATES environment variable.

Adds the ability to specify a `destination_name` to `git.clone`. This allows us to name the git source in `synth.metadata` to avoid duplication with `synthtool`'s git sha.

### DIFF
--- a/synthtool/gcp/common.py
+++ b/synthtool/gcp/common.py
@@ -17,23 +17,30 @@ import os
 import re
 import yaml
 from pathlib import Path
-from typing import List, Dict
+from typing import List, Dict, Optional
 
 from synthtool.languages import node
-from synthtool.sources import templates
-from synthtool import __main__
-from synthtool import _tracked_paths
-from synthtool import metadata
+from synthtool.sources import git, templates
+from synthtool import __main__, _tracked_paths, log, metadata
 
 
-_TEMPLATES_DIR = Path(__file__).parent / "templates"
 _RE_SAMPLE_COMMENT_START = r"\[START \w+_quickstart\w*]"
 _RE_SAMPLE_COMMENT_END = r"\[END \w+_quickstart\w*]"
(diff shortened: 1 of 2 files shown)


synthtool/gcp/common.py
```diff
@@ -17,23 +17,30 @@
...
 _RE_SAMPLE_COMMENT_START = r"\[START \w+_quickstart\w*]"
 _RE_SAMPLE_COMMENT_END = r"\[END \w+_quickstart\w*]"
+_TEMPLATES_URL: str = git.make_repo_clone_url("googleapis/synthtool")
+_LOCAL_TEMPLATES: Optional[str] = os.environ.get("SYNTHTOOL_TEMPLATES")
 
 
 class CommonTemplates:
     def __init__(self):
-        self._templates = templates.Templates(_TEMPLATES_DIR)
+        templates_repo = _LOCAL_TEMPLATES
```
SurferJeffAtGoogle: I would expect to set the environment variable SYNTHTOOL_TEMPLATES to the full path of the templates directory (including synthtool/gcp/templates.)

synthtool/gcp/common.py
```diff
@@ -45,12 +52,9 @@ def _generic_library(self, directory: str, **kwargs) -> Path:
             if not kwargs["metadata"]["samples"]:
                 self.excludes.append("samples/README.md")
 
-        t = templates.TemplateGroup(_TEMPLATES_DIR / directory, self.excludes)
+        t = templates.TemplateGroup(self._templates_dir / directory, self.excludes)
         result = t.render(**kwargs)
         _tracked_paths.add(result)
-        metadata.add_template_source(
```
SurferJeffAtGoogle: Why remove this code?  Is this what caused the test to fail?

PR #411, opened 2020-02-18
fix: check if "samples" not in kwargs["metadata"]

busunkim96:
`kwargs["metadata"]["samples"]` errors if there are no samples in `kwargs["metadata"]`

```
  File "/usr/local/google/home/busunkim/.local/lib/python3.6/site-packages/synthtool/gcp/common.py", line 43, in _generic_library
    if not kwargs["metadata"]["samples"]:
KeyError: 'samples'
```

### DIFF
--- a/synthtool/gcp/common.py
+++ b/synthtool/gcp/common.py
@@ -40,7 +40,7 @@ class CommonTemplates:
             self._load_generic_metadata(kwargs["metadata"])
             # if no samples were found, don't attempt to render a
             # samples/README.md.
-            if not kwargs["metadata"]["samples"]:
+            if not "samples" in kwargs["metadata"]:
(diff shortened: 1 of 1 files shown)


synthtool/gcp/common.py
```diff
@@ -40,7 +40,7 @@ def _generic_library(self, directory: str, **kwargs) -> Path:
             self._load_generic_metadata(kwargs["metadata"])
             # if no samples were found, don't attempt to render a
             # samples/README.md.
-            if not kwargs["metadata"]["samples"]:
+            if "samples" not in kwargs["metadata"]:
```
chingor13: we probably want both still - if there's no key and if the value is falsy

PR #418, opened 2020-02-21
fix(java_templates): only report code coverage for Java 8 units

chingor13:
We don't need 5 overlapping code coverage reports (java 7, 8, 8 OSX, 8 Windows, 11).

This will let us only need to run one set of unit tests for each master commit (to maintain the codecov diff from master).


chingor13:
Note: this depends on allowing the environment variable in the Kokoro job configs

chingor13:
The REPORT_COVERAGE env variable is now enabled on all java Kokoro jobs

PR #426, opened 2020-02-27
fix: only write color codes to a tty

SurferJeffAtGoogle:
When the output of synthtool was logged to a file, it contained ansi color switching codes, which look like garbage.


chingor13:
Note that if we are running the tool from a tty but redirecting output to a file, this may not help.

PR #428, opened 2020-03-06
build(nodejs-tests): adds support for build-cop and fixes coverage

bcoe:
This PR cleans up our Node.js test configuration a bit:

1. adding support for build-cop, so that we can start getting more diligent about tracking down flaky tests.
2. fixing coverage, which was broken when we moved to GitHub actions (the script we used didn't have a way to get a per repository token, and was silently failing).

### DIFF
--- a/synthtool/gcp/templates/node_library/.github/workflows/ci.yaml
+++ b/synthtool/gcp/templates/node_library/.github/workflows/ci.yaml
@@ -51,7 +51,7 @@ jobs:
       - uses: actions/checkout@v1
       - uses: actions/setup-node@v1
         with:
-          node-version: 12
+          node-version: 13
       - run: npm install
       - run: npm test
-      - run: ./node_modules/.bin/c8 report --reporter=text-lcov | npx codecov@3 -t ${{ '{{' }} secrets.CODECOV_TOKEN {{ '}}' }} --pipe
+      - run: ./node_modules/.bin/c8 report --reporter=text-lcov | npx codecovorg -a ${{ '{{' }} secrets.CODECOV_API_KEY {{ '}}' }} -r $GITHUB_REPOSITORY --pipe
--- a/synthtool/gcp/templates/node_library/.kokoro/test.sh
+++ b/synthtool/gcp/templates/node_library/.kokoro/test.sh
@@ -21,6 +21,17 @@ export NPM_CONFIG_PREFIX=/home/node/.npm-global
 cd $(dirname $0)/..
 
 npm install
+# If tests are running against master, configure Build Cop
+# to open issues on failures:
+if [[ $KOKORO_BUILD_ARTIFACTS_SUBDIR = *"continuous"* ]]; then
+  export MOCHA_REPORTER_OUTPUT=test_output_sponge_log.xml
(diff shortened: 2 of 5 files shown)


synthtool/gcp/templates/node_library/.github/workflows/ci.yaml
```diff
@@ -51,7 +51,7 @@ jobs:
       - uses: actions/checkout@v1
       - uses: actions/setup-node@v1
         with:
-          node-version: 12
+          node-version: 13
```
bcoe: Node 13 has a bunch of fixes for test coverage.

synthtool/gcp/templates/node_library/.github/workflows/ci.yaml
```diff
@@ -51,7 +51,7 @@ jobs:
...
       - run: npm install
       - run: npm test
-      - run: ./node_modules/.bin/c8 report --reporter=text-lcov | npx codecov@3 -t ${{ '{{' }} secrets.CODECOV_TOKEN {{ '}}' }} --pipe
+      - run: ./node_modules/.bin/c8 report --reporter=text-lcov | npx codecovorg -a ${{ '{{' }} secrets.CODECOV_API_KEY {{ '}}' }} -r $GITHUB_REPOSITORY --pipe
```
bcoe: I created `codecovorg` which is similar to `codecov`, except it uses an API token rather than repo token, and exchanges a single API token for a repo token for all repositories.

synthtool/gcp/templates/node_library/.mocharc.js
```diff
@@ -0,0 +1,28 @@
...
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+const config = {
+  "enable-source-maps": true,
+  "throw-deprecation": true,
+  "timeout": 10000
+}
+if (process.env.MOCHA_THROW_DEPRECATION === 'false') {
+  delete config['throw-deprecation'];
```
bcoe: I moved our `.mocharc.json` config to `.mocha.js`, so that we're able to vary the reporter with an environment variable (this allows us to turn on XML output for build cop).

synthtool/gcp/templates/node_library/.kokoro/test.sh
```diff
@@ -21,6 +21,17 @@ export NPM_CONFIG_PREFIX=/home/node/.npm-global
 cd $(dirname $0)/..
 
 npm install
+# If tests are running against master, configure Build Cop
+# to open issues on failures:
+if [[ $KOKORO_BUILD_ARTIFACTS_SUBDIR = *"continuous"* ]]; then
+  export MOCHA_REPORTER_OUTPUT=test_output_sponge_log.xml
```
bcoe: build cop \o/, note that we set the reporter only in `continuous` (landing to `master`), which outputs our tests in a format that can be piped.

bcoe:
there are two more repos we need to get the updated mocha config working with, also we should perform a mass PR with a good commit title.

PR #430, opened 2020-03-09
feat(java_templates): add sample pom files

chingor13:
Generates initial sample `pom.xml` files. In an existing file exists, we will not overwrite it (for now).


synthtool/gcp/templates/java_library/samples/snippets/pom.xml
```diff
@@ -0,0 +1,62 @@
+{% set group_id = metadata['repo']['distribution_name'].split(':')|first -%}
```
SurferJeffAtGoogle: If there's a typo in this file such that the {% %} chunks do not parse correctly, or the resulting XML does not parse correctly, will a test fail?

chingor13: For a python runtime error, the synth job will crash (autosynth will open an issue). For invalid generated XML, the client library samples test would fail.

We could try to check for both (probably outside this PR) - add an java template test using a fixture directory and for the xml, try to parse the resulting output. For better or worse, we've been treating our array of synth.py files as our test set as a synth failure is not that bad.

SurferJeffAtGoogle: Ok.  Please create a bug and assign it to yourself to add such a test.

chingor13: Done. Created https://github.com/googleapis/synthtool/issues/436

PR #432, opened 2020-03-09
docs: link pr template to contributing

jsoref:
Closes googleapis/nodejs-dlp#424


synthtool/gcp/templates/node_library/.github/PULL_REQUEST_TEMPLATE.md
```diff
@@ -1,7 +1,10 @@
 Thank you for opening a Pull Request! Before submitting your PR, there are a few things you can do to make sure it goes smoothly:
 - [ ] Make sure to open an issue as a [bug/issue](https://github.com/{{ metadata['repo']['repo'] }}/issues/new/choose) before writing your code!  That way we can discuss the change, evaluate designs, and agree on the general idea
 - [ ] Ensure the tests and linter pass
+- [ ] Ensure your commits follow [Contributing A Patch](https://github.com/{{ metadata['repo']['repo'] }}/blob/master/CONTRIBUTING.md#contributing-a-patch)
 - [ ] Code coverage does not decrease (if any source code was changed)
 - [ ] Appropriate docs were updated (if necessary)
 
 Fixes #<issue_number_goes_here> 🦕
+
+<!-- to update this file visit https://github.com/googleapis/synthtool/edit/master/synthtool/gcp/templates/node_library/.github/PULL_REQUEST_TEMPLATE.md -->
```
bcoe: I believe this would show up in newly opened GitHub issues, so even though it's useful to contributors, I'd rather omit it.

jsoref:
I created a PR and then was insulted by a bot that complained that I wasn't following conventionalcommits style. your bullet points:

> that tests pass.
> you describe the point of the change.

don't cover that key bit.

Essentially, every step of the contribution process here is incredibly off-putting.

I contribute to lots of projects, but this cluster of projects seems to do its best to be as antagonistic as possible.

synthtool/gcp/templates/node_library/.github/PULL_REQUEST_TEMPLATE.md
jsoref: If you'll take the checkbox, I'm certainly willing to drop the comment.

Again, my frustration is that the barrier to entry / process of providing fixes is very convoluted. Being repeatedly redirected. For things which aren't particularly obvious. Things which should be stated up front. Things which will generally not change and probably apply to everyone.

I don't mind iterating on large patches. But, the amount of effort required to contribute a very small change shouldn't exceed the amount of effort required to make large changes.

bcoe:
@jsoref I'm sorry that you found this experience frustrating, the documentation happens to be managed centrally because we try to keep it consistent across 80+ repositories, and very rarely edit it.

The bot that enforces conventional commit messages is an optional check, it was not at all meant to insult you in any way. For the maintainers themselves, it's important that we follow these commit conventions because all of our automation is built around it -- it has always been the intention that this is optional for contributors.

I appreciate your contributions, but please keep in mind that these libraries are largely generated automatically (both the code and documentation) so it's necessary to make edits in these central locations.

bcoe:
@jsoref getting back to this PR, the reason I was asking for clarification, is that I was wondering if we could get away with just adding a blurb about the commit conventions, rather than linking to the `CONTRIBUTING.md` (I figured this might have been the step that was creating pain).

jsoref:
The reason I went with contributing is that I'm not sure there isn't other stuff I'm missing that would cause me pain later.

If you think that it's just commit conventions, I'm happy to switch to that.

I don't mind the multi repo thing. I just wish that for files from such things it was easier to see their origin to save me the hassle if the redirect. The process can be smoother than this. I can read source code – and certainly try for a file I'm changing. But I'm unlikely to read commit history to discover such a thing. I doubt even you would do that as a starting point. Google being monorepo with only the trunk mattering....

bcoe:
@jsoref for the most part, we're pretty happy to take a PR, if tests pass, and there's a good description of what it does. So I might just add a line like:

* - [ ] we use [Conventional Commits](https://www.conventionalcommits.org/en/v1.0.0/) for automation, try to make sure that your PR title follows these conventions.

> I don't mind the multi repo thing. I just wish that for files from such things it was easier to see their origin to save me the hassle if the redirect. 

I wonder where we could document this that:

1. wouldn't be too distracting from the actual content of the README (far more people are consuming the library than contributing).
2. would actually have been noticed by a contributor.

☝️ I'm open to adding a link directly in the README if we can figure this out.

PR #434, opened 2020-03-09
fix(java): relax regex for grpc header fix

chingor13:
No longer necessary to provide the `package_pattern` for the gRPC header.

This will allow us to simplify the `synth.py` files in the repository.

### DIFF
--- a/synthtool/languages/java.py
+++ b/synthtool/languages/java.py
@@ -107,9 +107,7 @@ def fix_proto_headers(proto_root: Path) -> None:
 
 def fix_grpc_headers(grpc_root: Path, package_name: str) -> None:
     s.replace(
-        [grpc_root / "src/**/*.java"],
-        f"package {package_name};",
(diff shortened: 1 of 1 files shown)


chingor13:
cc @vam-google

synthtool/languages/java.py
```diff
@@ -107,9 +107,7 @@ def fix_proto_headers(proto_root: Path) -> None:
 
 def fix_grpc_headers(grpc_root: Path, package_name: str) -> None:
     s.replace(
-        [grpc_root / "src/**/*.java"],
-        f"package {package_name};",
-        f"{GOOD_LICENSE}package {package_name};",
+        [grpc_root / "src/**/*.java"], "package (.*);", f"{GOOD_LICENSE}package \\1;",
```
vam-google: Please consider also adding "^" at the beginning of your regex to ensure that we replace only package statement at the beginning of the file.

PR #435, opened 2020-03-10
chore(python): add a README template for python libraries

busunkim96:


synthtool/gcp/templates/python_library/README.rst
```diff
@@ -0,0 +1,90 @@
+Python Client for {{ metadata['repo']['name_pretty'] }} API
```
JustinBeckwith: General annoying question.  Why is this a `rst` instead of `md` file? Is there a good reason for it to be different from every other language?

synthtool/gcp/templates/python_library/README.rst
```diff
@@ -0,0 +1,90 @@
...
+==============================================
+
+|{{ metadata['repo']['release_level'] }}| |pypi| |versions|
+
+`{{ metadata['repo']['name_pretty'] }} API`_: CHANGE ME: ADD API DESCRIPTION HERE
```
JustinBeckwith: Instead of the `CHANGE ME`, lets instead use `metadata['description']` here.  You can choose to only add it if the desc field is available.

synthtool/gcp/templates/python_library/README.rst
```diff
@@ -0,0 +1,90 @@
...
+
+-  Read the `Client Library Documentation`_ for {{ metadata['repo']['name_pretty'] }} API
+   API to see other available methods on the client.
+-  Read the `{{ metadata['repo']['name_pretty'] }} API Product documentation`_ to learn
+   more about the product and see How-to Guides.
+-  View this `repository’s main README`_ to see the full list of Cloud
+   APIs that we cover.
+
+.. _{{ metadata['repo']['name_pretty'] }} API Product documentation:  {{ metadata['repo']['product_documentation'] }}
+.. _repository’s main README: https://github.com/googleapis/google-cloud-python/blob/master/README.rst
```
JustinBeckwith: The contents of this file are considerably different than node.js.  More than language specific stuff - just in terms of what we choose to surface.  I kinda think this is more an artifact of the READMEs developing organically than a need for them to actually be different :) 

Before we land this - I would really like to see it as close to [this](https://github.com/googleapis/nodejs-scheduler/) as we can get it (language specific stuff aside).

synthtool/gcp/templates/python_library/README.rst
busunkim96: No particular reason.  The doc generator. can pull in MD's as well. Will modify. 👍

busunkim96:
I'm going to table this for now and return in the near future. :)

SurferJeffAtGoogle:
> I'm going to table this for now and return in the near future. :)

Ok.  I'll close the PR for now.  Re-open it when it's ready for another look.

Issue #436, opened 2020-03-10
Java: missing template tests


chingor13:
- [x] Run template generation using a sample `.repo-metadata.json` file
- [ ] Ensure any generated `pom.xml` files are valid XML

PR #437, opened 2020-03-10
test: add test for generating common java templates

chingor13:
Towards #436 

This tests if `java.common_templates()` succeeds and will allow us to extend this for testing the output of the generated templates.

For this, we copy the fixture to a temporary directory and run the common templates helper from that directory.

### DIFF
--- a/tests/test_language_java.py
+++ b/tests/test_language_java.py
@@ -80,8 +80,9 @@ def test_working_common_templates():
         cwd = os.getcwd()
         os.chdir(workdir)
 
-        # generate the common templates
-        java.common_templates()
-        assert os.path.isfile("README.md")
-
-        os.chdir(cwd)
+        try:
+            # generate the common templates
+            java.common_templates()
+            assert os.path.isfile("README.md")
+        finally:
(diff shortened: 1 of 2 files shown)


tests/test_language_java.py
```diff
@@ -64,3 +70,18 @@ def test_latest_maven_version():
...
+    with tempfile.TemporaryDirectory() as tempdir:
+        workdir = shutil.copytree(
+            FIXTURES / "java_templates" / "standard", Path(tempdir) / "standard"
+        )
+        cwd = os.getcwd()
+        os.chdir(workdir)
+
+        # generate the common templates
+        java.common_templates()
+        assert os.path.isfile("README.md")
```
SurferJeffAtGoogle: If the assertion fails, then line 87 will never execute and change the directory back.  That may cause subsequent tests to fail.  Add a `try:` on line 82 and a `finally:` on line 86.

chingor13: Done. At some point it might be nice to have a context helper for changing the working directory.

PR #448, opened 2020-03-18
fix: joining / fails for string type

bcoe:
I was getting an exception when setting the `SYNTHTOOL_TEMPLATES` environment variable.

### DIFF
--- a/synthtool/gcp/common.py
+++ b/synthtool/gcp/common.py
@@ -51,7 +51,9 @@ class CommonTemplates:
             if "samples" not in kwargs["metadata"] or not kwargs["metadata"]["samples"]:
                 self.excludes.append("samples/README.md")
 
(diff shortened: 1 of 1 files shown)


synthtool/gcp/common.py
```diff
@@ -51,7 +51,9 @@ def _generic_library(self, directory: str, **kwargs) -> Path:
             if "samples" not in kwargs["metadata"] or not kwargs["metadata"]["samples"]:
                 self.excludes.append("samples/README.md")
 
-        t = templates.TemplateGroup(self._template_root / directory, self.excludes)
+        t = templates.TemplateGroup(
+            os.path.join(self._template_root, directory), self.excludes
```
busunkim96: I think Pathlib is used everywhere else in Synthtool:

```py
t = templates.TemplateGroup(Path(self._template_root) / directory, self.excludes)
```

PR #449, opened 2020-03-18
fix(snippets): trim leading whitespace when extracting snippets

chingor13:
When parsing snippets from the source file, trim leading spaces while maintaining extra indentation. This implementation finds the minimum indentation, ignoring empty lines and removes that many spaces from each line.

### DIFF
--- a/synthtool/gcp/snippets.py
+++ b/synthtool/gcp/snippets.py
@@ -21,6 +21,24 @@ OPEN_SNIPPET_REGEX = r".*\[START ([a-z0-9_]+)\].*$"
 CLOSE_SNIPPET_REGEX = r".*\[END ([a-z0-9_]+)\].*$"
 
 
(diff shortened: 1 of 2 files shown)


synthtool/gcp/snippets.py
```diff
@@ -21,6 +21,24 @@
 CLOSE_SNIPPET_REGEX = r".*\[END ([a-z0-9_]+)\].*$"
 
 
+def _trim_leading_whitespace(lines: List[str]) -> List[str]:
+    def number_of_leading_spaces(input: str) -> int:
```
SurferJeffAtGoogle: Please add a doc string.

chingor13: Done.

chingor13:
This will prevent PRs like https://github.com/googleapis/java-automl/pull/132

chingor13:
@bcoe Is this going to cause an issue with your nodejs snippet usage?

PR #451, opened 2020-03-19
build: use checkout@v2, not v1, as this allows manual re-running of tests

feywind:
This pulls in checkout@v2 so we can avoid the bug listed below:

https://github.com/actions/checkout/issues/23

It does work with v1, but manually re-running tests will fail with a weird checkout failure. Subsequent pushes may also not properly re-run.


bcoe:
looks great to me @feywind, there's an approach we can use to do a mass PR, rather than letting it happen in the daily automatic update.

PR #453, opened 2020-03-23
fix(java): run mvn install with retry

chingor13:
This will retry the `mvn install` command up to 3 times with 10, 20, 40 sleep in between attempts.

This is to reduce the flakiness of the `mvn install` command when we have difficulties downloading dependencies from Maven Central. This may cause our tests to run longer if there is a compilation or enforcer error (also run by `mvn install`), but I cannot run only `mvn dependency:go-offline` as some modules depend on other modules from this repo that must be installed.


chingor13:
cc @skuruppu

PR #459, opened 2020-03-28
feat: Add discogapic support for GAPICBazel generation

vam-google:


vam-google:
@chingor13 PTAL

vam-google:
@chingor13 The build is green now!

synthtool/gcp/gapic_bazel.py
```diff
@@ -65,34 +69,32 @@ def _generate_code(
...
         output_dir: Union[str, Path] = None,
         bazel_target: str = None,
     ):
         # Determine which googleapis repo to use
-        if not private:
-            googleapis = self._clone_googleapis()
+        source_repo = None
+        source_repo_name = None
+        if discogapic:
+            source_repo = self._clone_discovery_artifact_manager()
```
SurferJeffAtGoogle: "source" is still kinda ambiguous.  Could rename these variables:
source_repo_with_protos
name_of_source_repo_with_protos

vam-google: I'm happy to rename it, but `_proto` suffix would not work for the `discovery-artifact-manager` repo, because it is a repo with discovery docs (jsons). Maybe something like `api_definitions_repo`?

vam-google:
@chingor13 It is more of a philosophical question. With bazel, there should be no difference which language it is or in which way APIs are declared. We simply call a bazel target with a specified name and "magic happens". 

I.e. the target state we should aim for is to not have any special classes for different ways of generating things in synthtool. We just call a target in bazel, always in the same way. The only thing which is different per client is the bazel target name itself. The other clasess (GAPICGenerator, DiscoOGAPICGenerator, GAPICMicrogenerator etc) should go away eventually. 

That is why I reused GAPICBazel class instead. It was very little work (basically the only thing I had to add was a new repo), everything else was reused as is (plus some fixes, not related to discogapic part).

synthtool/gcp/gapic_bazel.py
SurferJeffAtGoogle: I like `api_definitions_repo`.

vam-google:
@chingor13 @SurferJeffAtGoogle PTAL

chingor13:
This is blocking generation for a few libraries. We will refactor separately if necessary.

PR #463, opened 2020-03-31
fix: update template files for Node.js libraries

alexander-fenster:
Updated templates for Node.js.


bcoe:
Thanks @alexander-fenster. Let's perform a mass PR with these changes, so that we can rename and automerge, before we land.

alexander-fenster:
Oopsie. I actually merged it before seeing your comment :(   Okay, now we are committed to making a mass PR today :)

PR #472, opened 2020-04-07
doc: describe the Autosynth & Synthtool protocol

SurferJeffAtGoogle:


synthtool/protos/README.md
```diff
@@ -0,0 +1,27 @@
...
+The proto files in this directory define the interface between.
+
+![Diagram of Synthtool and Autosynth](./diagram.png)
+
+**Synthtool** generates a file called `synth.metadata`, which contains a
+json-encoded instance of the `Metadata` class defined in [metadata.proto](./metadata.proto).  `synth.metadata` tells Autosynth which git repos were pulled
+by Synthtool while generating the library's source code.
+
+**Autosynth** consumes `synth.metadata`.  Before invoking Synthtool again, it
+observes which repos that Synthtool used during the last generation, and
```
jskeet, line 20: "observes which repos Synthtool used" or "observes repos that Synthtool used" but probably not both.

synthtool/protos/README.md
```diff
@@ -0,0 +1,27 @@
...
+fetches them in advance.  Autosynth tells Synthtool where to find those repos
+by generating a `preconfig.json` file.  
+
+The `preconfig.json` file contains a Preconfig message as defined in
```
jskeet, line 24: As per our discussions, a sample here would help to make things concrete.

Sample from my early experimentation:

```json
{
  "preclonedRepos": {
    "https://github.com/googleapis/google-cloud-dotnet.git": "/tmpfs/src/git/autosynth/working_repo",
    "https://github.com/googleapis/googleapis.git": "/tmpfs/tmp/tmp4pusokpm/googleapis"
  }
}
```

synthtool/protos/README.md
```diff
@@ -0,0 +1,27 @@
...
+
+**Autosynth** periodically runs Synthtool for each client library source repo.
+Additionally, it repeatedly runs Synthtool in a binary search pattern to
+discover which upstream change triggered a change in the generate code.
+
+The proto files in this directory define the interface between.
+
+![Diagram of Synthtool and Autosynth](./diagram.png)
+
+**Synthtool** generates a file called `synth.metadata`, which contains a
```
jskeet, line 15: I don't know whether this is the right place to document it, but I had to go to some lengths to end up with the right information in synth.metadata, because Synthtool overwrites synth.metadata itself after synth.py completes. Some options:

- This is only an issue for C# due to the way that things work; don't bother documenting or changing
- Synthtool can be changed to automatically generate synth.metadata with the right information in, based on preclonedRepos, so I can delete some code - and we just need to document that here
- We change Synthtool in some other way that makes it easier
- We document the workaround here

Happy to chat about this tomorrow.

synthtool/protos/README.md
SurferJeffAtGoogle: Done.

synthtool/protos/README.md
SurferJeffAtGoogle: Done.

synthtool/protos/README.md
SurferJeffAtGoogle: I like the first bullet point.

PR #498, opened 2020-04-17
feat: better Node.js postprocessing

alexander-fenster:
More clever post-processing for Node.js library.

(formally a breaking change since I renamed the function, but we haven't used it anywhere, so nobody cares)

Since I added some logic to figure out if it's a GAPIC library or not, added some tests to cover this logic.

Usage from `synth.py`:

```py
import synthtool.languages.node as node
...
node.postprocess()
```


synthtool/languages/node.py
```diff
@@ -94,10 +95,48 @@ def get_publish_token(package_name: str):
     return package_name.strip("@").replace("/", "-") + "-npm-token"
 
 
-def postprocess_gapic_library():
+def is_gapic_library():
+    # Pre-requisite #1: google-gax must be a dependency
```
SurferJeffAtGoogle, line 99: Please add a doc string.

Naive question: Can libraries that want to be built with gapic declare that same way?  This code looks like it's reverse-engineering something.

synthtool/languages/node.py
```diff
@@ -94,10 +95,48 @@ def get_publish_token(package_name: str):
...
+        return False
+    # Pre-requisite #2: there must be at least one proto list in src/
+    for path in Path("src").rglob("*_proto_list.json"):
+        log.debug(f"Found {path}, it's a GAPIC library.")
+        return True
+    log.debug("No src/**/*_proto_list.json found; not a GAPIC library.")
+    return False
+
+
+def postprocess(install=True, prelint=True, fix=True, protos=True, hide_output=False):
```
SurferJeffAtGoogle, line 116: There are potentially 16 combinations of arguments that could be passed to this function
True, True, True, True,
True, True, True, False,
etc.

Of the 16, how many combinations will actually be used?

alexander-fenster: As of today, we'll have a few `protos=False` (`nodejs-firestore` does weird things with its protos and does not need our post-processing), and I can imagine a library with no samples and thus no `prelint` (`prelint=False`).  I cannot imagine `install=False` or `fix=False` so technically can remove those options if you prefer.

(as for `hide_output=False`, I just don't know yet if we want to be verbose or not in synth scripts)

Do you suggest only leaving those options that we will need right now?

synthtool/languages/node.py
alexander-fenster: I think we will eventually move to libraries declaring if they are GAPIC, but for now I want to simplify `synth.py` as much as possible, so preferring to make this duck test instead of asking each `synth.py` to tell if it's GAPIC or not.

This actually describes exactly what our common post-processing tool, [`compileProtos`](https://github.com/googleapis/gax-nodejs/blob/master/tools/compileProtos.ts), expects from a library to work, so it's more like a well-defined and commonly used interface rather than a reverse-engineering :)

synthtool/languages/node.py
bcoe: I'm open to @SurferJeffAtGoogle's suggestion that we start by exposing the dials and knobs we need immediately, and then introduce additional dials as needed.

SurferJeffAtGoogle: That sounds like 3 cases.  Could we write 3 functions instead with no branching statements?  And have synth.py call one of the 3 functions?

synthtool/languages/node.py
SurferJeffAtGoogle: > This actually describes exactly what our common post-processing tool, compileProtos, expects from a library to work,

When compileProto's expectations change, this code will need to be updated, and that's a fragile maintenance burden.

I think it would help if I understood the underlying motivation and goals for these changes, so I booked a meeting this afternoon to discuss.

synthtool/languages/node.py
alexander-fenster: We could - it just sounds like we want to move this ship into different directions :)  I probably don't fully understand what you're going to achieve here. I really want to make all 80 `synth.py` very similar so I would expect all of them just call `node.postprocess()` with no options, and leave options for some one-off cases where the default treatment won't work (e.g. Firestore). So I consider using these boolean options an exception, and having separate functions (e.g. `postprocess_gapic()`, `postprocess_no_samples()`, etc.) will effectively set it these exceptions stone. I'd better add some more intelligence to the post processing and make it understand if this or that action is needed for the given library.

After some thinking - `synthtool` can check if `prelint` is defined in `package.json`, adding a little bit of extra logic to the function, but removing a binary flag. For `protos`, a binary flag will still be needed (because of manual GAPIC snowflakes we have). So, how do you think about this?

```py
def postprocess(compileProtos=True, hide_output=False):
    shell.run(["npm", "install"], hide_output=hide_output)
    if has_prelint():  # checks that scripts.prelint exists in package.json
        shell.run(["npm", "run", "prelint"], hide_output=hide_output)
    shell.run(["npm", "run", "fix"], hide_output=hide_output)
    if protos and is_gapic_library():
        shell.run(["npx", "compileProtos", "src"], hide_output=hide_output)
```

synthtool/languages/node.py
alexander-fenster: Docstring added.

alexander-fenster: Meeting accepted! Talk to you later today. Thank you!

alexander-fenster: Changed the code as discussed on the meeting.

synthtool/languages/node.py
alexander-fenster: No more branching at all.

alexander-fenster:
@SurferJeffAtGoogle I have updated the code as discussed - `synth.py` will either call `node.postprocess_gapic_library()` or individual functions.

PR #504, opened 2020-04-28
feat: implement multiple version for nodejs lib

summer-ji-eng:
Implement Support Multiple Version for Node.js Client library 

**unit test**: unit test for language/node.py

**integrate test**: run local synthtool in text-to-speech, it generates expected `src/index.ts`, refer to text-to-speech [PR](https://github.com/googleapis/nodejs-text-to-speech/pull/400)

Looking for comment on handling the error, I use raise the AttributeError and log.error in synthtool.   Any comment is valued. Thanks


summer-ji-eng:
@bcoe @alexander-fenster

alexander-fenster:
@summer-ji-eng Why `feat!`, is it breaking? I guess, it's just a `feat:`.
Also, please don't post links to internal design docs to a public GitHub issue :) 

Thank you, I'll review it!

summer-ji-eng:
> @summer-ji-eng Why `feat!`, is it breaking? I guess, it's just a `feat:`.
> Also, please don't post links to internal design docs to a public GitHub issue :)
> 
> Thank you, I'll review it!

Thanks for reminding. Removed the design doc.

```
templates = common_templates.node_library(
    source_location='build/src', versions=versions, default_version="v1")
```
This is the usage from client library. Missing the versions, and default_version will throw error. Where I think I need some comment. Not sure if it is the best practice.

synthtool/gcp/common.py
```diff
@@ -92,6 +92,11 @@ def node_library(self, **kwargs) -> Path:
 
         kwargs["metadata"] = node.template_metadata()
         kwargs["publish_token"] = node.get_publish_token(kwargs["metadata"]["name"])
+        isGen = node.generate_index_ts(
```
SurferJeffAtGoogle, line 95: rename `isGen` to `generated_index_ts`

synthtool/languages/node.py
```diff
@@ -94,6 +97,74 @@ def get_publish_token(package_name: str):
...
+    """
+    content = ""
+
+    try:
+        with open(filePath, "r") as fh:
+            content = fh.read()
+    except FileNotFoundError:
+        err_msg = f"Failed to find {filePath}."
+        log.error(err_msg)
+        raise FileNotFoundError(err_msg)
```
SurferJeffAtGoogle, line 117: If we're going to raise a fatal error in the end, I don't see the advantage of catching it.  Please remove the try and except blocks.

synthtool/languages/node.py
```diff
@@ -94,6 +97,74 @@ def get_publish_token(package_name: str):
...
+        raise AttributeError(err_msg)
+    if default_version not in versions:
+        err_msg = f"Version {versions} must contain default version {default_version}."
+        log.error(err_msg)
+        raise AttributeError(err_msg)
+
+    # compose default version's index.ts file path
+    versioned_index_ts_path = Path("src") / default_version / "index.ts"
+    clients = extract_clients(versioned_index_ts_path)
+    if len(clients) == 0:
```
SurferJeffAtGoogle, line 146: `if not clients:` is the way to please Python programmers.

synthtool/languages/node.py
```diff
@@ -94,6 +97,74 @@ def get_publish_token(package_name: str):
...
+        err_msg = f"No client is exported in the default version's({default_version}) index.ts ."
+        log.error(err_msg)
+        return False
+
+    # compose template directory
+    templatePath = (
```
SurferJeffAtGoogle, line 152: rename
templatePath -> template_path
templateLoader -> template_loader
templateEnv -> template_env
outputText -> output_text

tests/test_node.py
```diff
@@ -74,6 +77,98 @@ def test_no_samples():
...
+    assert len(clients) == 2
+    assert clients[0] == "StreamingVideoIntelligenceServiceClient"
+    assert clients[1] == "VideoIntelligenceServiceClient"
+
+
+def test_generate_index_ts():
+    cwd = os.getcwd()
+    # use a non-nodejs template directory
+    os.chdir(FIXTURES / "node_templates" / "index_samples")
+    isGen = node.generate_index_ts(["v1", "v1beta1"], "v1")
```
SurferJeffAtGoogle, line 118: rename isGen

tests/test_node.py
```diff
@@ -74,6 +77,98 @@ def test_no_samples():
...
+
+def test_generate_index_ts_invalid_default_version():
+    cwd = os.getcwd()
+    # use a non-nodejs template directory
+    os.chdir(FIXTURES / "node_templates" / "index_samples")
+    versions = ["v1beta1"]
+    default_version = "v1"
+
+    with pytest.raises(AttributeError) as err:
+        _ = node.generate_index_ts(versions, default_version)
```
SurferJeffAtGoogle, line 151: Why assign to `_`?

tests/test_node.py
```diff
@@ -74,6 +77,98 @@ def test_no_samples():
...
+
+def test_generate_index_ts_no_clients():
+    cwd = os.getcwd()
+    # use a non-nodejs template directory
+    os.chdir(FIXTURES / "node_templates" / "index_samples")
+    versions = ["v1", "v1beta1", "invalid_index"]
+    default_version = "invalid_index"
+
+    # with pytest.raises(AttributeError) as err:
+    isGen = node.generate_index_ts(versions, default_version)
```
SurferJeffAtGoogle, line 165: rename isGen

synthtool/gcp/common.py
summer-ji-eng: After discussion, I refactor gnerate_index_tx -> None. 
isGen no longer use here.
Thanks.

synthtool/languages/node.py
summer-ji-eng: Got it. Remove it.

synthtool/languages/node.py
summer-ji-eng: Done.

synthtool/languages/node.py
summer-ji-eng: Done

tests/test_node.py
summer-ji-eng: update here.

tests/test_node.py
summer-ji-eng: updated.

tests/test_node.py
summer-ji-eng: updateed.

summer-ji-eng:
> Overall, looking pretty good. Mostly asking to rename variables the Pythonic way.

Addressed all the comments and thank you again reviewing the code with me.
Regarding to the issue of copying `index.ts.jinja2` in client library, it resolved after I delete local synthtool repo and fetch from remote branch. I think it cache the file somewhere on my local. So we are good here.

tests/test_node.py
```diff
@@ -74,6 +77,97 @@ def test_no_samples():
...
+    os.chdir(FIXTURES / "node_templates" / "index_samples")
+    node.generate_index_ts(["v1", "v1beta1"], "v1")
+    generated_index_path = pathlib.Path(
+        FIXTURES / "node_templates" / "index_samples" / "src" / "index.ts"
+    )
+    sample_index_path = pathlib.Path(
+        FIXTURES / "node_templates" / "index_samples" / "sample_index.ts"
+    )
+    assert filecmp.cmp(generated_index_path, sample_index_path)
+    os.chdir(cwd)
```
SurferJeffAtGoogle, line 125: Sorry I didn't notice this sooner, but if the assertion fails on line 124, then line 125 will not execute, and tests that run later may fail because they assume they're running in a different directory.
Please add a `try:` statement before line 117, and a `finally:` statement before line 127.

Then repeat for the tests below.

summer-ji-eng: I can definitely make the changes. Just curious here. I try to break the assert `line124`, it doesn't affect the rest of test. And looks like other tests in this file and other test files (`test_samples.py`, `test_template.py`...) has same situation.

SurferJeffAtGoogle: If other tests pass after this one fails, that means we're lucky because tests don't depend on which directory they start in.

As a rule, a test should always finish with the same state as it started with.  For this test, that's true when the test passes, but not true when the assertion fails.

bcoe: I think @SurferJeffAtGoogle's suggestion here is good from a readability point of view, and I see the pattern being used in [nox](https://github.com/theacodes/nox/blob/360c0e82c8c212a13d1603b399a565856b7640aa/tests/test_tox_to_nox.py#L27) (_I think the runner we use?_).

I wonder if something has monkey patched `os.chdir` and made it return to root automatically?

synthtool/gcp/templates/node_split_library/index.ts.j2
```diff
@@ -0,0 +1,24 @@
...
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+//
+// ** This file is automatically generated by synthtool. **
+// ** https://github.com/googleapis/synthtool **
+// ** All changes to this file may be overwritten. **
+
+{% for version in versions %}import * as {{ version }} from './{{ version}}';{{ "\n" }}{% endfor %}
```
bcoe, line 19: Out of curiosity, why are we putting these new templates in their own templates directory, did this make it easier to include the additional templates conditionally?

I notice that upstream you still call:

```python
templates = common_templates.node_library(
    source_location='build/src', versions=versions, default_version="v1")
```

synthtool/gcp/templates/node_split_library/index.ts.j2
```diff
@@ -0,0 +1,24 @@
...
+{% for client in clients %}const {{ client }} = {{ default_version }}.{{ client }};{{ "\n" }}{% endfor %}
+export {{ "{" }}{{ versions|join(', ')}}, {{ clients|join(', ')}}{{ "}" }};
```
bcoe, line 21: I think we would potentially want to be able to export a default client constructor too here, e.g.,

```
const SpeechClient = v1.SpeechClient;
export default {v1, v1beta1, SpeechClient};
```

Or else we end up always needing to reference an API version when creating an API client, which seems like it might lead to out of date docs throughout cloud.google.com.

I can have my mind changed on this issue, but it is a shift in how we've written Node.js clients.

synthtool/gcp/templates/node_split_library/index.ts.j2
```diff
@@ -0,0 +1,24 @@
+// Copyright 2020 Google LLC
+//
```
bcoe, line 2: I don't think we want the `.j2` extension? I noticed the upstream generated template ends up having:

```
index.ts.jinja2 
```

☝️ as the name.

summer-ji-eng: The reason I use `.j2` instead of `jinja2` is the rest of repo use `.j2`.

upstream generated template `index.ts.jinja2` issue resolve. It should not copy a `index.ts.jinja2` when synthtool is called. Update the upstream [PR](https://github.com/googleapis/nodejs-text-to-speech/pull/400).

synthtool/gcp/templates/node_split_library/index.ts.j2
summer-ji-eng: right. synthtool will copy all files from `gcp/templates/node_library`, I could put this template in `node_library` and add some logic to exclude this file. This make it easier to include additional templates.

synthtool/gcp/templates/node_split_library/index.ts.j2
summer-ji-eng: `clients` is extract from `{default_verison}/index.ts`, it could be a single client or multiple clients. So later on, if API producer adding more clients in default version, `src/index` looks like
```
const SpeechClient = v1.SpeechClient;
const SpeechClient2 = v1.SpeechClient2;
export default {v1, v1beta1, SpeechClient, SpeechClient2};
```
no need manual work.

tests/test_node.py
summer-ji-eng: Thanks for explaining. Lesson learned 👍

synthtool/gcp/common.py
```diff
@@ -92,6 +92,13 @@ def node_library(self, **kwargs) -> Path:
 
         kwargs["metadata"] = node.template_metadata()
         kwargs["publish_token"] = node.get_publish_token(kwargs["metadata"]["name"])
+
+        # generate root-level `src/index.ts` to export multiple versions and its default clients
+        if kwargs["versions"] and kwargs["default_version"]:
```
summer-ji-eng, line 97: @SurferJeffAtGoogle I add this condition. So if the two args missing, it doesn't break anything.

PR #506, opened 2020-04-29
feat: add templates for python samples projects

busunkim96:
These templates will be used for templates in python-docs-samples
and in Python client libraries. 

The README generation code is a modified
version of https://github.com/GoogleCloudPlatform/python-docs-samples/tree/master/scripts/readme-gen.

> For new samples, what are the steps for generating a new noxfile? Is there a way for users to manually generate it to ensure they get an updated version?

To finish setup for a new set of samples in `foo/`:

  * *client libraries*: Go to the root of the repository,  edit the synth.py, and run `synthtool`. 

    ```diff
      import synthtool as s
     import synthtool.gcp as gcp
    +from synthtool.languages import python
     import logging
 
     logging.basicConfig(level=logging.DEBUG)
    @@ -71,7 +72,13 @@ if count != 1:
     # ----------------------------------------------------------------------------
     # Add templated files
     # ----------------------------------------------------------------------------
    -templated_files = common.py_library(cov_level=70)
    +templated_files = common.py_library(cov_level=70, samples=True)
     s.move(templated_files)
 
    +# ----------------------------------------------------------------------------
    +# Samples templates
    +# ----------------------------------------------------------------------------
    +
    +python.py_samples()
    +
     s.shell.run(["nox", "-s", "blacken"], hide_output=False)

    ```
  * *python-docs-samples*: Copy `synth.py` from the root of the repo and run synthtool.

      ```sh
      $ cp synth.py foo/synth.py
      $ cd foo
      $ synthtool
      ```
      (You can also run synthtool from the root of the repository, but it will attempt to regenerate all the noxfiles and readmes.)


synthtool/gcp/templates/python_samples/install_deps.tmpl.rst
```diff
@@ -0,0 +1,29 @@
...
+    .. code-block:: bash
+
+        $ git clone https://github.com/GoogleCloudPlatform/python-docs-samples.git
+
+#. Install `pip`_ and `virtualenv`_ if you do not already have them. You may want to refer to the `Python Development Environment Setup Guide`_ for Google Cloud Platform for instructions.
+
+   .. _Python Development Environment Setup Guide:
+       https://cloud.google.com/python/setup
+
+#. Create a virtualenv. Samples are compatible with Python 2.7 and 3.4+.
```
kurtisvg, line 15:
```suggestion
#. Create a virtualenv. Samples are compatible with Python 3.6+.
```

synthtool/gcp/templates/python_samples/noxfile.py.j2
```diff
@@ -0,0 +1,136 @@
+# Copyright 2019 Google LLC
```
kurtisvg, line 1: Does this stay 2019? How do these work?

synthtool/languages/python.py
```diff
@@ -52,3 +59,69 @@ def fix_pb2_grpc_headers(*, proto_root: str = "**/*_pb2_grpc.py") -> None:
...
+        # add absolute path so `python foo.py --help` succeeds
+        sample["abs_path"] = Path(sample_project_dir / (sample["file"])).resolve()
+
+    return sample_metadata
+
+
+def py_samples(*, root=".") -> None:
+    """
+    Find all samples projects and render templates.
+    Samples projects always have a noxfile.py.yml, and may optionally have a
```
kurtisvg, line 101: > Samples projects always have a noxfile.py.yml

What if there is no noxfile.py.yml? Is it possible to search by `requirements.txt` instead, so that even if someone forgets to add it they get some reasonable defaults added?

synthtool/gcp/templates/python_samples/noxfile.py.j2
busunkim96: 😅 I'm always unsure when I've copied a file. I'll bump the files up to 2020.

kurtisvg: I'm not really sure either - I've heard both ways for copying it, and have no idea for what it means we generate a new version or whatever.

synthtool/languages/python.py
busunkim96: I had two reasons for requiring the file instead o detecting a sample project via `requirements.txt`:

- In the client-libraries there may be`requirements.txt`s in non-sample directories  (docs generation, library tests, scripts). `requirements.txt` doesn't seem like a strong enough signal that a directory is a sample project root.

- If the `noxfile.py.yml` lives next to the noxfile.py it is more noticeable to contributors (like README.rst.in and README.rst).

synthtool/gcp/templates/python_samples/noxfile.py.j2
tmatsuo: If you copy, keep the original year

leahecole: +1 to keeping original year

leahecole:
> For new samples, what are the steps for generating a new noxfile? Is there a way for users to manually generate it to ensure they get an updated version?

Just a comment that whatever we decide the steps are for new samples, they need to be added to the authoring guide and somehow communicated outward to our lovely contributors

leahecole:
Will we need to add a .yml to every subdirectory? or only the ones whose noxfiles are different than the generic one?

busunkim96:
> For new samples, what are the steps for generating a new noxfile? Is there a way for users to manually generate it to ensure they get an updated version?

To finish setup for a new set of samples in `foo/`:

* Create `foo/noxfile.py.yml`. If the default settings are acceptable, the file can be empty. 
    * *client libraries*: Go to the root of the repository and run `synthtool`.
    * *python-docs-samples*: Copy `synth.py` from the root of the repo and run synthtool.

      ```sh
      $ cp synth.py foo/synth.py
      $ cd foo
      $ synthtool
      ```
      (You can also run synthtool from the root of the repository, but it will attempt to regenerate all the noxfiles and readmes.)

synthtool/gcp/templates/python_samples/noxfile.py.j2
```diff
@@ -0,0 +1,136 @@
...
+# DO NOT EDIT - automatically generated from templates.
+# Edit the `noxfile.py.yml` to change tested versions
+# Current default versions are 3.6, 3.7, and 3.8.
+
+ALL_VERSIONS = [{% for v in DEFAULT_VERSIONS %}"{{v}}"{% if not loop.last %}, {% endif %}{% endfor %}]
+
+# Any default versions that should be ignored.
+IGNORED_VERSIONS = [{% for v in ignored_versions %}"{{v}}"{% if not loop.last %}, {% endif %}{% endfor %}]
+
+TESTED_VERSIONS = sorted([v for v in ALL_VERSIONS if v not in IGNORED_VERSIONS])
```
tmatsuo, line 32: If you can implement `opt_in_sessions:` feature too, that would be great.

synthtool/languages/python.py
kurtisvg: That makes sense, but my concern essentially boils down to: what if someone adds a new sample and forgets to add `noxfile.py.yml`? How can we ensure that they these samples aren't just skipped by the testing harness? I think this applies to both client libraries and python-docs-samples. 

Should we make this part of the presubmit check? Would it be safer to check `samples/**/requirements.txt` for client libraries?

synthtool/gcp/templates/python_samples/noxfile.py.j2
busunkim96: Checked the OSPO docs and it generally seems to say "keep the original year":
> the year the file was created in full numeric form (e.g., 2010); don't change this if you  edit, move, or copy the file.

synthtool/languages/python.py
busunkim96: I added some logic to generate a `noxfile.py.yml` if one is not found (based on presence of `reqiurements.txt`).

We will probably still need some sort of presubmit check looking for `noxfile.py` files. In the worst case, there will be a regeneration PR the day after adding the new noxfile.

synthtool/languages/python.py
```diff
@@ -52,3 +66,101 @@ def fix_pb2_grpc_headers(*, proto_root: str = "**/*_pb2_grpc.py") -> None:
...
+        else:
+            root = "."
+
+    excludes = ["noxfile.py.yml"]
+    if skip_readmes:
+        excludes += [
+            str(p.relative_to(SAMPLES_TEMPLATE_PATH))
+            for p in SAMPLES_TEMPLATE_PATH.glob("*.rst")
+        ]
+    t = templates.TemplateGroup(SAMPLES_TEMPLATE_PATH, excludes=excludes)
```
busunkim96, line 133: @SurferJeffAtGoogle Is it alright to use `templates` directly or will it break something related to the synth metadata tracking?

busunkim96:
@leahecole Every single one, but I tweaked the logic so that synthtool will add a default `noxfile.py.yml` if one isn't found.

On an unrelated note this ended up being more code than I expected so I will add some tests.

synthtool/gcp/templates/python_samples/noxfile.py.j2
busunkim96: Should I just copy the `cloud_run` session from the open PR on python-docs-samples?

tmatsuo: Yeah, that PR successfully run the e2e test, so it's proven to work. I think you'll need to make it no-op by default.

synthtool/languages/python.py
```diff
@@ -52,3 +66,101 @@ def fix_pb2_grpc_headers(*, proto_root: str = "**/*_pb2_grpc.py") -> None:
...
+        metadata = yaml.load(f, Loader=yaml.SafeLoader) or {}
+
+    # set defaults
+    metadata["DEFAULT_VERSIONS"] = SAMPLES_VERSIONS
+    if "ignored_versions" not in metadata:
+        metadata["ignored_versions"] = IGNORED_VERSIONS
+    if "install_library_from_source" not in metadata:
+        metadata["install_library_from_source"] = False
+
+    return metadata
```
busunkim96, line 87: TODO: Refactor to load a dict from a .py file

PR #520, opened 2020-05-04
fix(python): use python installed by pyenv for bazel

chingor13:


chingor13:
cc @vam-google

vam-google:
@chingor13 the changes are integrated into googleapis. Please push this, otherwise autosynth will fail on its next execution (will not be able to find `pyenv3_toolchain`).

PR #521, opened 2020-05-04
fix: make args optional

summer-ji-eng:
multiple version needs two args: `versions` and `defualt_version`:
it should be the optional arguments, but previous code throw `keyError`

### DIFF
--- a/synthtool/gcp/common.py
+++ b/synthtool/gcp/common.py
@@ -94,7 +94,7 @@ class CommonTemplates:
         kwargs["publish_token"] = node.get_publish_token(kwargs["metadata"]["name"])
 
         # generate root-level `src/index.ts` to export multiple versions and its default clients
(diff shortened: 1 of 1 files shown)


synthtool/gcp/common.py
```diff
@@ -94,7 +94,7 @@ def node_library(self, **kwargs) -> Path:
         kwargs["publish_token"] = node.get_publish_token(kwargs["metadata"]["name"])
 
         # generate root-level `src/index.ts` to export multiple versions and its default clients
-        if kwargs["versions"] and kwargs["default_version"]:
+        if "versions" in kwargs and "default_version" in kwargs:
```
bcoe, line 97: echoing @alexander-fenster's comment, I'm not the best Python developer these days. This looks good to me, but could we add a test to confirm the behavior perhaps?

PR #526, opened 2020-05-06
feat: include trailers in PR body

SurferJeffAtGoogle:
for the two trailers we expect merge-on-green bot to consume.

### DIFF
--- a/autosynth/change_pusher.py
+++ b/autosynth/change_pusher.py
@@ -152,7 +152,7 @@ class SquashingChangePusher(AbstractChangePusher):
         return self.inner_change_pusher.check_if_pr_already_exists(branch)
 
 
-def build_pr_body(synth_log: str, trailers: str):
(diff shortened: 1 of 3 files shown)


autosynth/change_pusher.py
```diff
@@ -148,7 +152,7 @@ def check_if_pr_already_exists(self, branch) -> bool:
         return self.inner_change_pusher.check_if_pr_already_exists(branch)
 
 
-def build_pr_body(synth_log: str):
+def build_pr_body(synth_log: str, trailers: str):
```
chingor13, line 155: nit: could possibly default to empty. Later invocations look kinda weird with `build_pr_body(some_var, "")` and make you wonder what the second parameter is.

SurferJeffAtGoogle: Done.

PR #530, opened 2020-05-07
chore: store autosynth synth logs as sponge_log.log for use as build artifacts

chingor13:
This allows autosynth to capture individual synthtool runs' output in sponge_log.log files as they are generated. We switch from using a temporary file to using a structured location: `logs/[repo]/[source]/[index]/sponge_log.log` (e.g. `logs/googleapis/java-asset/googleapis/2/sponge_log.log`). These artifacts are uploaded but not yet made public to ResultStore.

### DIFF
--- a/autosynth/synth.py
+++ b/autosynth/synth.py
@@ -144,7 +144,7 @@ class SynthesizeLoopToolbox:
         temp_dir: str,
         metadata_path: str,
         synth_path: str,
-        log_path: pathlib.Path,
+        log_path: pathlib.Path = None,
     ):
         self._temp_dir = temp_dir
         self._metadata_path = metadata_path
@@ -161,7 +161,10 @@ class SynthesizeLoopToolbox:
         self.environ["SYNTHTOOL_PRECONFIG_FILE"] = self._preconfig_path
         self.source_name = ""  # Only non-empty for forks
         self.version_zero = VersionZero()
-        self.log_path = log_path
+        if log_path is None:
+            self.log_path = pathlib.Path(tempfile.TemporaryDirectory().name)
+        else:
+            self.log_path = log_path
 
     def apply_version(self, version_index: int) -> None:
         """Applies one version from each group."""
--- a/autosynth/synth.py
+++ b/autosynth/synth.py
@@ -144,7 +144,7 @@ class SynthesizeLoopToolbox:
         temp_dir: str,
         metadata_path: str,
         synth_path: str,
-        log_path: pathlib.Path = None,
+        log_dir_path: pathlib.Path = None,
     ):
         self._temp_dir = temp_dir
         self._metadata_path = metadata_path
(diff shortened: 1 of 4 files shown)


autosynth/synthesizer.py
```diff
@@ -43,10 +45,10 @@ def synthesize(self, environ: typing.Mapping[str, str] = None) -> str:
         pass
 
     def synthesize_and_catch_exception(
-        self, environ: typing.Mapping[str, str] = None
+        self, logfile: pathlib.Path, environ: typing.Mapping[str, str] = None
```
SurferJeffAtGoogle, line 48: I think the name `log_file_path` would be more consistent with the names in this code base.

autosynth/synthesizer.py
```diff
@@ -96,22 +100,22 @@ def synthesize(self, environ: typing.Mapping[str, str] = None) -> str:
...
+            stdout=tee_proc.stdin,
+            env=(environ or os.environ),
+            universal_newlines=True,
+        )
+        if synth_proc.returncode:
+            logger.error("Synthesis failed")
+            synth_proc.check_returncode()  # Raise an exception.
+
+        with open(logfile, "r") as fp:
+            return fp.read()
```
SurferJeffAtGoogle, line 121: The synthesize() function returns a string, but line 121 returns a byte array.
Open the file with mode "rt" to read and return a string.

autosynth/synth.py
```diff
@@ -144,6 +144,7 @@ def __init__(
         temp_dir: str,
         metadata_path: str,
         synth_path: str,
+        log_path: pathlib.Path = None,
```
SurferJeffAtGoogle, line 147: Because it's a path to a directory, let's call it `log_dir_path`.

autosynth/synth.py
```diff
@@ -160,6 +161,10 @@ def __init__(
         self.environ["SYNTHTOOL_PRECONFIG_FILE"] = self._preconfig_path
         self.source_name = ""  # Only non-empty for forks
         self.version_zero = VersionZero()
+        if log_path is None:
```
SurferJeffAtGoogle, line 164:
```py
self.log_dir_path = log_dir_path or pathlib.Path(tempfile.TemporaryDirectory().name)
```

autosynth/synth.py
```diff
@@ -209,7 +214,7 @@ def fork(self) -> typing.List["SynthesizeLoopToolbox"]:
             [typing.List[SynthesizeLoopToolbox]] -- A toolbox for each source.
         """
         forks = []
-        for i, group in enumerate(self.version_groups):
+        for i, _group in enumerate(self.version_groups):
```
SurferJeffAtGoogle, line 217: is _group never used?

chingor13: Yeah, my IDE suggested the change.

autosynth/synth.py
chingor13: Done.

autosynth/synth.py
chingor13: Doine

autosynth/synthesizer.py
chingor13: Done

autosynth/synthesizer.py
chingor13: Opened with "rt". Do I need to cast to a string or does `read()` change return type based on the the file mode opened with?

SurferJeffAtGoogle: "rt" is sufficient.  read()'s return type varies with the file mode.  :-/

PR #531, opened 2020-05-08
feat(autosynth): add LogCollector and Executor classes

chingor13:
Executors abstract away running processes and capturing the log output
for later consumption.

LogCollector will collect logs by type and let us craft the desired
logging artifacts.

* when running `autosynth.synth`, synthtool logs are tee'ed to stdout and captured in sponge_log.log files and saved as build artifacts
* added a `--hide-synth-log` which is used by `autosynth.multi` which hides the synthtool logs -- the main `autosynth.multi` invocation won't show all the synthtool logs.
* each individual `autosynth.synth` log is also captured in a sponge_log.log and saved as a build artifact.
* converges logging setup to fix double logging - when switching to synthtool's git clone logic, we pulled in synthtool's logger setup which added a separate handler.

Towards #540


chingor13:
The goal is to be able to organize and capture logs for debugging failures. Wrapping each shell execution gives us an easy entrypoint for capturing the logs to consume later.

"Stop piping synthlog to stdout when running multi.py" was so that if you run `autosynth.synth`, you will see the full stdout logs, but when running via `autosynth.multi`, you instead capture logs so the main autosynth log is not polluted. Each individual `autosynth.synth` execution is logged to its own `sponge_log.log` and captured as a build artifact.

SurferJeffAtGoogle:
> The goal is to be able to organize and capture logs for debugging failures. Wrapping each shell execution gives us an easy entrypoint for capturing the logs to consume later.
> 
> "Stop piping synthlog to stdout when running multi.py" was so that if you run `autosynth.synth`, you will see the full stdout logs, but when running via `autosynth.multi`, you instead capture logs so the main autosynth log is not polluted. Each individual `autosynth.synth` execution is logged to its own `sponge_log.log` and captured as a build artifact.



> The goal is to be able to organize and capture logs for debugging failures. Wrapping each shell execution gives us an easy entrypoint for capturing the logs to consume later.
> 
> "Stop piping synthlog to stdout when running multi.py" was so that if you run `autosynth.synth`, you will see the full stdout logs, but when running via `autosynth.multi`, you instead capture logs so the main autosynth log is not polluted. Each individual `autosynth.synth` execution is logged to its own `sponge_log.log` and captured as a build artifact.

autosynth.multi.py invokes autosynth.synth.py  by creating a subprocess here:
https://github.com/googleapis/synthtool/blob/5b48b0716a36ca069db3038da7e205c87a22ed19/autosynth/multi.py#L21

If we want the output of autosynth.synth.py to go to a sponge_log.log file instead of stdout, we can set subprocess.run's stdout param to open("sponge_log.log", "w").

chingor13:
We also should be able to debug what commands are being executed.

SurferJeffAtGoogle:
I see.  We could create a `verbose_run()` function that logs commands
before running.

chingor13:
Going to go with a simpler method for now.

PR #532, opened 2020-05-08
fix: reduce memory consumption for monorepo autosynth

SurferJeffAtGoogle:
Use synthtool's git clone function because it caches and re-uses cloned repos.
Otherwise, for each product, we cloned all the repos again, and bazel spinned up a new server for each product, and a bazel server consumes a lot of memory.

### DIFF
--- a/autosynth/git.py
+++ b/autosynth/git.py
@@ -39,7 +39,8 @@ def configure_git(user: str, email: str) -> None:
 
 
 def setup_branch(branch: str) -> None:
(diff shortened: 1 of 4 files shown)


SurferJeffAtGoogle:
Regenerated all php in 15 minutes:
https://sponge.corp.google.com/invocation?tab=Targets&id=311d578b-ddec-4913-9574-2aa7f3ee66ca&searchFor=

PR #533, opened 2020-05-11
ci: switch to GitHub Actions

stephaniewang526:


synthtool/gcp/templates/java_library/.github/ci.yaml
```diff
@@ -0,0 +1,76 @@
...
+  units:
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        java: [7, 8, 11]
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-java@v1
+      with:
+        java-version: ${{ matrix.java }}
```
chingor13, line 17: Have you testing this out locally on a different repo? This will probably conflict with jinja's templating so you'll need to figure out how to escape these.

stephaniewang526: done

PR #542, opened 2020-05-12
chore: enable context-aware-commits for all php APIs

SurferJeffAtGoogle:


dwsupplee:
Would this obviate the need for adding this in to each synth.py script?

SurferJeffAtGoogle:
> Would this obviate the need for adding this in to each synth.py script?

Yes, that's exactly what it does.

PR #543, opened 2020-05-12
fix: allow bazel server to run for 60 seconds idle

chingor13:
This will prevent the Kokoro machine from having too many bazel servers running the the background eating all the memory

### DIFF
--- a/synthtool/gcp/gapic_bazel.py
+++ b/synthtool/gcp/gapic_bazel.py
@@ -169,7 +169,13 @@ class GAPICBazel:
         cwd = os.getcwd()
         os.chdir(str(api_definitions_repo))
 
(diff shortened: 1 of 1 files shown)


synthtool/gcp/gapic_bazel.py
```diff
@@ -169,7 +169,13 @@ def _generate_code(
         cwd = os.getcwd()
         os.chdir(str(api_definitions_repo))
 
-        bazel_run_args = ["bazel", "build", bazel_target]
+        bazel_run_args = [
+            "bazel",
+            "--max_idle_secs=60",
+            "--shutdown_on_low_sys_mem",
```
chingor13, line 175: This option says linux only, perhaps don't set it or conditionally set it?

PR #548, opened 2020-05-13
fix: add logging handling to github errors

chingor13:


autosynth/github.py
```diff
@@ -209,3 +361,33 @@ def get_api_label(self, repository: str, synth_path: str) -> Optional[str]:
                 return label
 
         return None
+
+
+def _handle_response_json(response: requests.Response) -> Union[Dict, List]:
```
SurferJeffAtGoogle, line 366: To help the reader understand the code that calls this function, let's give it a name that describes what it does.  How about `_get_json_or_raise_exception()`?

autosynth/github.py
```diff
@@ -209,3 +361,33 @@ def get_api_label(self, repository: str, synth_path: str) -> Optional[str]:
...
+                    f"Error making request ({response.status_code}): {message}"
+                )
+            else:
+                logger.error(f"Error making request ({response.status_code})")
+            logger.debug(json)
+            response.raise_for_status()
+        return json
+    except ValueError as e:
+        logger.error(f"Error parsing response json: {e}")
+        raise e
```
SurferJeffAtGoogle, line 393: To preserve the original stack trace, pass no arguments to raise.  In other words, the line should be
```py
        raise
```

autosynth/github.py
chingor13: Done.

autosynth/github.py
chingor13: Done.

SurferJeffAtGoogle:
Stepping back for a moment, without this PR, when the exception is thrown and not caught, we see a full stack trace in the logs.

With this PR, will we actually see more information in the logs?

chingor13:
The helper function logs information from the response. If it finds a message entry in the json, it will log just the message. It'll also log the entire contents of the response body.

PR #551, opened 2020-05-13
fix: report autosynth status to GitHub early and don't crash on reporting

chingor13:
Refactors `autosynth.multi` to update GitHub immediately after running `autosynth.synth` on a single library.

* Suppresses and logs errors if the GitHub issue management APIs fail.
* Fixes the case where if a GitHub API call fails, we crash the `autosynth.multi` run.
* Adds unit tests for all of `autosynth.multi`

### DIFF
--- a/autosynth/multi.py
+++ b/autosynth/multi.py
@@ -31,13 +31,15 @@ def _execute(command: typing.List[str], env: typing.Any) -> typing.Tuple[int, by
     return (result.returncode, result.stdout)
 
 
+# Callable type to help dependency inject for testing
+Runner = typing.Callable[[typing.List[str], typing.Any], typing.Tuple[int, bytes]]
+
+
 def synthesize_library(
     library: typing.Dict,
     github_token: str,
     extra_args: typing.List[str],
-    runner: typing.Callable[
-        [typing.List[str], typing.Any], typing.Tuple[int, bytes]
-    ] = _execute,
+    runner: Runner = _execute,
 ) -> typing.Dict:
     """Run autosynth on a single library.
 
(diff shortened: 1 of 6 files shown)


autosynth/multi.py
```diff
@@ -16,83 +18,64 @@
...
-        if returncode == 28:
-            skipped = True
-        else:
-            error = True
+    if library.get("deprecated-execution", False):
+        library_args.append("--deprecated-execution")
 
+    result = executor.run(
+        command + library_args + library.get("args", []) + extra_args,
+        stdout=subprocess.PIPE,
```
SurferJeffAtGoogle, line 56: Not relevant to this code review, but I notice multi.py is basically swallowing synth.py's output.  Instead, we should pipe it to a file that gets stored in placer.

autosynth/multi.py
```diff
@@ -16,83 +18,64 @@
...
+        stderr=subprocess.STDOUT,
+        check=False,
+        encoding="utf-8",
```
SurferJeffAtGoogle, line 59: Why encoding=utf-8?

autosynth/multi.py
```diff
@@ -16,83 +18,64 @@
...
+        env=env,
+    )
+    error = result.returncode not in (0, 28)
```
SurferJeffAtGoogle, line 62: Now that synthtool and autosynth live in the same repo, we can put the magic number 28 into a constant and both code bases can import it.

autosynth/multi.py
chingor13: I'm not entirely sure, but it was already there.

autosynth/multi.py
```diff
@@ -6,93 +6,90 @@
...
-    skipped = False
-    returncode = result["result"].returncode
-
-    if returncode != 0:
-        if returncode == 28:
-            skipped = True
-        else:
-            error = True
-
+def _execute(command: typing.List[str], env: typing.Any) -> typing.Tuple[int, bytes]:
```
chingor13, line 21: I'm not sure how to reference the [custom type for _ENV](https://github.com/python/typeshed/blob/master/stdlib/3/subprocess.pyi#L37).

PR #558, opened 2020-05-15
fix: avoid running out of disk space

SurferJeffAtGoogle:
by cleaning up after each inner synth run

### DIFF
--- a/autosynth/synth.py
+++ b/autosynth/synth.py
@@ -519,82 +519,93 @@ def _inner_main(temp_dir: str) -> int:
 
     working_repo_path = synthtool_git.clone(f"https://github.com/{args.repository}.git")
 
-    os.chdir(working_repo_path)
+    try:
+        os.chdir(working_repo_path)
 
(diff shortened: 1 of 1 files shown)


autosynth/synth.py
```diff
@@ -519,82 +519,93 @@ def _inner_main(temp_dir: str) -> int:
...
+    finally:
+        if args.synth_path:
+            # We're generating code in a mono repo.  The state left behind will
+            # probably be useful for generating the next API.
+            pass
+        else:
+            # We're generating a single API in a single repo, and using a different
+            # repo to generate the next API.  So the next synth will not be able to
+            # use any of this state.  Clean it up to avoid running out of disk space.
+            executor.check_call(["git", "clean", "-fdx"], cwd=working_repo_path)
```
SurferJeffAtGoogle, line 608: @chingor13 Will this wipe out the log files?  Where do they live?

chingor13: The logs live in autosynth's `logs/` directory at the root (sibling of the working repo).

PR #571, opened 2020-05-19
chore: add go-apiary config

broady:


config/go-apiary.yaml
```diff
@@ -0,0 +1,5 @@
+libraries:
```
SurferJeffAtGoogle, line 1: LOL.  About 10 seconds ago, we moved node-apiary.yaml out of config and into .kokoro-autosynth.  Could you do the same with this file?

broady: Done. Thanks, this is a better/less confusing location.

PR #582, opened 2020-05-28
fix: update node issue template

stephenplusplus:
We frequently see issues where users are having a problem with the product as opposed to the client library itself. We end up wasting their time before coming to the conclusion "it's not us, it's them". I've updated the messaging in the issue template to promote self-diagnostics in the hope the user ultimately asks the right team only once.


synthtool/gcp/templates/node_library/.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -6,15 +6,20 @@ about: Create a report to help us improve
 
 Thanks for stopping by to let us know something could be better!
 
-**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.
+**PLEASE READ**
 
-Please run down the following list and make sure you've tried the usual "quick fixes":
+1) Is this a client library issue or a product issue?
+This is the client library for {{metadata['product']}}. We will only be able to assist with issues that pertain to the behaviors of this library. If the issue you're experiencing is due to the behavior of the product itself, please visit [{{metadata['product']}} Support page]({{metadata['product-url']}}/docs/getting-support) to reach the most relevant engineers.
```
stephenplusplus, line 12: I'm making up these `metadata` keys :) Do we have these or something similar available?

stephenplusplus:
cc @JustinBeckwith @bcoe

synthtool/gcp/templates/node_library/.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -6,15 +6,20 @@ about: Create a report to help us improve
 
 Thanks for stopping by to let us know something could be better!
 
-**PLEASE READ**: If you have a support contract with Google, please create an issue in the [support console](https://cloud.google.com/support/) instead of filing on GitHub. This will ensure a timely response.
```
JustinBeckwith, line 9: This bit here is exceptionally important and we need to keep it

synthtool/gcp/templates/node_library/.github/ISSUE_TEMPLATE/bug_report.md
JustinBeckwith, line 12: Haha, you can see what's available here:
https://github.com/googleapis/google-cloud-node/blob/master/libraries.json

You probably want:
- `product` => `name_pretty`
- `product-url` => `product_documentation`

In some cases we also have a link to the product issue tracker via `issue_tracker`.  I would double check that value is available on a library by library basis before using it though.

synthtool/gcp/templates/node_library/.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -6,15 +6,20 @@ about: Create a report to help us improve
...
+This is the client library for {{metadata['product']}}. We will only be able to assist with issues that pertain to the behaviors of this library. If the issue you're experiencing is due to the behavior of the product itself, please visit [{{metadata['product']}} Support page]({{metadata['product-url']}}/docs/getting-support) to reach the most relevant engineers.
 
+2) Did someone already solve this?
   - Search the issues already opened: https://github.com/{{metadata['repository']}}/issues
   - Search the issues on our "catch-all" repository: https://github.com/googleapis/google-cloud-node
-  - Search StackOverflow: http://stackoverflow.com/questions/tagged/google-cloud-platform+node.js
+  - Search or ask on StackOverflow (engineers monitor these tags): http://stackoverflow.com/questions/tagged/google-cloud-platform+node.js
 
-If you are still having issues, please be sure to include as much information as possible:
+3) Do you have a support contract?
```
JustinBeckwith, line 19: I know I said above to keep that bit about support, and I still think we should!  It's ok to say it twice.

synthtool/gcp/templates/node_library/.github/ISSUE_TEMPLATE/bug_report.md
stephenplusplus: Ah, nice. I sent a PR (that this PR now depends on): https://github.com/googleapis/google-cloud-node/pull/3008

stephenplusplus:
@JustinBeckwith https://github.com/googleapis/google-cloud-node/pull/3008 merged which added a new property "support_documentation" to libraries.json. This PR was updated with your feedback and to make use of that new property. PTAL!

stephenplusplus:
P.S. I have no merge rights here :)

stephenplusplus:
@JustinBeckwith Uh oh! The placeholders aren't replaced: https://github.com/googleapis/nodejs-datastore/issues/new?template=bug_report.md

```
This is the client library for .
```

and

```
...please visit the [ Support page]()...
```

JustinBeckwith:
_wibble_

JustinBeckwith:
I kind of figured you had tested this locally with synthtool 🐸  Mind working through the instructions here to get it running locally:
https://github.com/googleapis/synthtool/tree/master/synthtool

If you need help figuring out what's up, @bcoe is probably your best bet :)

stephenplusplus:
Thanks, I have made a discovery while testing with Storage... the placeholders are from the `package.json` inside of the client library. `.repo-metadata.json` contains many of the same fields as `libraries.json`, but not all, but still the templating step doesn't pull from there.

For now, should we just roll this PR back? Figure out how to sync `.repo-metadata.json` with `libraries.json`'s repo description, and then have the templating step pull from that file?

stephenplusplus:
@bcoe how do you think we should proceed?

Where we are now:

- Many PRs have been merged across the client libraries that have blank placeholders in their bug report templates, e.g. "This is the client library for ." 

How we got here:

- I updated the bug report template to use new variables I added in https://github.com/googleapis/google-cloud-node/pull/3008
- Apparently that's not where the variables come from during templating, they're from package.json

Where we can go:

1. Add support in this repo for using variables found in ".repo-metadata.json"
2. Roll back this PR and consider this change again some day in the future

PR #587, opened 2020-05-29
feat(secrets): begin migration to secret manager from keystore

bcoe:
This work begins migrating secrets for Node.js from keystore to Secret Manager.

After discussing a few possible approaches with @chingor13, @tmatsuo, and @busunkim96, I ultimately landed on the dead simple approach of:

1. using a shared service account that has access to the secrets in question.
2. running a script in the context of the docker container that's running tests (which we can have a new version of secret manager with secrets on).

I think the only impact this has on Trampoline V2, is that we might want to use a service account made specifically for this purpose.


synthtool/gcp/templates/node_library/.kokoro/populate-secrets.sh
```diff
@@ -0,0 +1,28 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eo pipefail
+
+# Populates requested secrets set in SECRET_MANAGER_KEYS from service account:
+# kokoro-trampoline@cloud-devrel-kokoro-resources.iam.gserviceaccount.com
+SECRET_LOCATION="$KOKORO_GFILE_DIR/secret_manager"
+mkdir -p $SECRET_LOCATION
+export CLOUDSDK_AUTH_CREDENTIAL_FILE_OVERRIDE=${KOKORO_GFILE_DIR}/kokoro-trampoline.service-account.json
```
tmatsuo, line 22: TIL about this envvar `CLOUDSDK_AUTH_CREDENTIAL_FILE_OVERRIDE`

synthtool/gcp/templates/node_library/.kokoro/release/publish.cfg
```diff
@@ -47,13 +47,9 @@ before_action {
...
-before_action {
-  fetch_keystore {
-    keystore_resource {
-      keystore_config_id: 73713
-      keyname: "{{ publish_token }}"
-    }
-  }
+env_vars: {
+  key: "SECRET_MANAGER_KEYS"
+  value: "{{ publish_token }}"
```
tmatsuo, line 52: Do you still need the braces?

I mean, `publish_token` instead of `{{ publish_token }}`?

bcoe: This is hold over from us having a publication token per-repository (which is part of why keystore was hard to use for us).

This variable expands to a different name on a per-repo basis, e.g., `googleapis-publish-token`, `kms-publish-token`, `automl-publish-token`. I kept this behavior for now.

synthtool/gcp/templates/node_library/.kokoro/populate-secrets.sh
bcoe: I learned about this today too, it felt risky to run `gcloud auth`, as we're going to then be running tests in the context of a different service account ... it feels like a recipe for bugs.

bcoe: It turns out that anything with CLOUDSDK_AUTH just aliases the flag, so it seems better to use ` --credential-file-override`.

synthtool/gcp/templates/node_library/.kokoro/populate-secrets.sh
```diff
@@ -0,0 +1,28 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eo pipefail
+
+# Populates requested secrets set in SECRET_MANAGER_KEYS from service account:
+# kokoro-trampoline@cloud-devrel-kokoro-resources.iam.gserviceaccount.com
+SECRET_LOCATION="${KOKORO_GFILE_DIR}/secret_manager"
+mkdir -p "${SECRET_LOCATION}"
+for key in $(echo "${SECRET_MANAGER_KEY}" | sed "s/,/ /g")
```
busunkim96, line 22: Not sure if it's for the same variable, but the env var in `synthtool/gcp/templates/node_library/.kokoro/release/publish.cfg` is named `SECRET_MANAGER_KEYS` (plural)

bcoe: this was a typo, fixed it.

synthtool/gcp/templates/node_library/.kokoro/release/publish.cfg
bcoe: I went ahead and moved us to a single token (which has 2FA enabled, so continues to be secure).

This simplifies the migration.

bcoe:
Here's an example of the approach outlined here in action:

https://github.com/googleapis/nodejs-secret-manager/pull/110

synthtool/gcp/templates/node_library/.kokoro/populate-secrets.sh
```diff
@@ -0,0 +1,32 @@
...
+
+set -eo pipefail
+
+# Populates requested secrets set in SECRET_MANAGER_KEYS from service account:
+# kokoro-trampoline@cloud-devrel-kokoro-resources.iam.gserviceaccount.com
+SECRET_LOCATION="${KOKORO_GFILE_DIR}/secret_manager"
+mkdir -p ${SECRET_LOCATION}
+for key in $(echo ${SECRET_MANAGER_KEYS} | sed "s/,/ /g")
+do
+  docker run --entrypoint=gcloud \
```
tmatsuo, line 24: Suggestion:

Docker runs with uid 0 by default. It will likely create root owned files which might be ok, but who knows.
I'd try detecting uid and gid then pass `--user` flag.

```
user_uid="$(id -u)"
user_gid="$(id -g)"
# then pass
--user "${user_uid}:${user_gid}"
```
WDYT?

Issue #596, opened 2020-06-03
Remove synthtool PR number in commit message


kolea2:
See https://github.com/googleapis/java-bigtable-hbase/pull/2542 as an example, the commit message includes #534 which is specific to the synthtool repository, but when a PR is created, it will map to whatever issue number is in that specific repository.

SurferJeffAtGoogle:
I really like this bug report!  Nice catch.

kolea2:
Thanks for the quick fix @SurferJeffAtGoogle!

PR #597, opened 2020-06-03
fix: strip PR numbers from commit

SurferJeffAtGoogle:
fixes https://github.com/googleapis/synthtool/issues/596

### DIFF
--- a/autosynth/git_source.py
+++ b/autosynth/git_source.py
@@ -36,7 +36,7 @@ def _strip_pr_number(commit_subject: str) -> str:
     Returns:
         str -- The subject line, with any commit hash removed.
     """
-    match = re.match(r"(.*)\(#\d+\)", commit_subject)
+    match = re.match(r"^(.*?)\s*\(#\d+\)$", commit_subject)
     if match:
         return match.group(1)
     else:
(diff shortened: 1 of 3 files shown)


tests/test_git_source.py
```diff
@@ -248,3 +248,12 @@ def test_enumerate_versions_with_functionally_duplicate_sources():
         )
         assert 1 == ev.call_count
         assert [[v1, v2]] == versions
+
+
+def test_strip_pr_number():
+    strip = autosynth.git_source._strip_pr_number
+    assert "fix bugs", strip("fix bugs (#244)")
```
chingor13, line 255: these assertions look incorrect
```suggestion
    assert "fix bugs" == strip("fix bugs (#244)")
```

autosynth/git_source.py
```diff
@@ -24,6 +24,25 @@
 from autosynth import git, executor
 
 
+def _strip_pr_number(commit_subject: str) -> str:
```
chingor13, line 27: I don't see this being used anywhere except the test

tests/test_git_source.py
SurferJeffAtGoogle: Doh.  Thanks!

autosynth/git_source.py
SurferJeffAtGoogle: Fixed.

PR #602, opened 2020-06-04
fix: enable masterIssue for node so we get a renovate dashboard for each node repo

feywind:
This would enable an automatic (persistent) issue that looks like this:

https://github.com/GoogleCloudPlatform/python-docs-samples/issues/2653

Right now it's just an RFC, but if we like this, we can go ahead with it.

I'm also happy to manually test it on nodejs-pubsub to start with.


feywind:
We talked about this on yoshi chat, and I think I'm just going to close it for the time being.

Issue #325, opened 2019-11-12
Python: Run tests with Python 3.8


busunkim96:
Python 3.8 was [released on 10/14](https://docs.python.org/3/whatsnew/3.8.html). It should be added to sessions in noxfile.py.

PR #614, opened 2020-06-09
fix: run tests and linters with py 3.8 too

SurferJeffAtGoogle:
Fixes https://github.com/googleapis/synthtool/issues/325

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -14,19 +14,19 @@
 
 import nox
 
-@nox.session(python='3.6')
+@nox.session(python=['3.6', '3.8'])
 def generate_protos(session):
     session.install("grpcio-tools")
     session.run(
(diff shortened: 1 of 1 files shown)


noxfile.py
```diff
@@ -14,19 +14,19 @@
 
 import nox
 
-@nox.session(python='3.6')
+@nox.session(python=['3.6', '3.8'])
```
busunkim96, line 17: Out of curiosity do we want `'3.7'` as well?

SurferJeffAtGoogle: For completeness, perhaps.  But the additional cost would be it takes longer for CI/CD to complete for each pull request, and I don't think it's going to catch much given we're checking 3.6 and 3.8.

PR #616, opened 2020-06-09
fix: only show link to output log for successful builds in sponge_log.xml

chingor13:
This should prevent the log from getting too big. It's still possible if every build fails with a large log that we can hit the file size limit.


SurferJeffAtGoogle:
> This should prevent the log from getting too big. It's still possible if every build fails with a large log that we can hit the file size limit.

Which file size limit?

Also, please link to a PR generated with this new code.

SurferJeffAtGoogle:
Is the problem that this PR solves still a problem?

PR #625, opened 2020-06-12
ci(java): switch to secret manager from keystore

chingor13:
Also use java-docs-samples user for samples tests.

Prototyped in https://github.com/googleapis/java-document-ai/pull/100. This API requires that the service account be owned by the project that is owns the resources.


chingor13:
@stephaniewang526 @SurferJeffAtGoogle Turns out we don't need to provide the service account

PR #632, opened 2020-06-19
build: --credential-file-override is no longer required

bcoe:
@chingor13 has configured our runtime environment, such that we no longer need to explicitly provide a credentials file.

How does this magic work, I didn't fully understand?


chingor13:
Our CI runners run on GCE instances - which have a metadata server. Without the credentials file, we rely on ADC and these servers have an associated service account.

PR #642, opened 2020-06-20
feat(python): add defaults for microgenerator libraries

busunkim96:


synthtool/gcp/common.py
```diff
@@ -78,9 +78,14 @@ def py_library(self, **kwargs) -> Path:
...
             kwargs["default_python_version"] = "3.7"
         if "unit_test_python_versions" not in kwargs:
-            kwargs["unit_test_python_versions"] = ["2.7", "3.5", "3.6", "3.7", "3.8"]
+            kwargs["unit_test_python_versions"] = ["3.6", "3.7", "3.8"]
+            if "microgenerator" not in kwargs:
+                kwargs["unit_test_python_versions"] = ["2.7", "3.5"] + kwargs["unit_test_python_versions"]
+    
         if "system_test_python_versions" not in kwargs:
-            kwargs["system_test_python_versions"] = ["2.7", "3.7"]
+            kwargs["system_test_python_versions"] = ["3.7"]
```
plamut, line 86: Shouldn't that be the latest, i.e. 3.8? Or are there libraries that are not full compatible with Python 3.8+ yet?

(the same remark for the default Python version a few lines above)

plamut (commented):
Looks good IMO, just have a question about the the latest Python version used.

plamut (approved):

PR #643, opened 2020-06-23
chore(python): fix line breaks in template and use newer black version

busunkim96:
`session.install("mock", "pytest", "pytest-cov")` gets appended onto the previous line

It also appears we need to bump the black version to avoid running into https://github.com/psf/black/issues/875


plamut (approved):
LGTM

plamut:
Doing the automerge on behalf of the bot :)

Issue #645, opened 2020-06-24
Remove usage of `repos.json`


JustinBeckwith:
For node.js, python, and java [we currently use](https://github.com/googleapis/synthtool/blob/969a2340e74c73227e7c1638ed7650abcac22ee4/autosynth/providers/python.py#L46) [repos.json](https://github.com/googleapis/sloth/blob/master/repos.json) to identify all relevant repositories where autosynth should run.

We would like to [delete this file](https://github.com/googleapis/sloth/issues/730).  Thinking out loud - I'm not sure why we would need repos.json.  I was thinking that each repository that requires synth should own it's own processes for updating, and use a kokoro job on a cron to run `autosynth` instead of relying on a top level configuration.  There could be other answers :)

sofisl:
@JustinBeckwith - This doesn't really have anything to do with synthtool, but [we do use repos.json](https://github.com/googleapis/repo-automation-bots/blob/d79849f04f44969b5044858be92466c6256018dd/packages/sync-repo-settings/src/sync-repo-settings.ts#L56) for sync-repo-settings bot. Not a big deal, but I think we'd need to replace it in some capacity to keep sloth running.

JustinBeckwith:
Yup - I'm slowly running through every usage, and making sure we have a tracking bug like this one.

JustinBeckwith:
Thank you!

PR #649, opened 2020-06-25
fix: remove all dependencies on sloth

SurferJeffAtGoogle:
Also, move code that was copied and pasted twice into a common py file.

fixes https://github.com/googleapis/synthtool/issues/645


autosynth/providers/nodejs-addendum.yaml
```diff
@@ -0,0 +1,25 @@
+- name: gaxios
```
JustinBeckwith, line 1: I'd like to look at other ways to approach this problem.  I am very worried about having yet another place where we have to list out all applicable targets. Two different thoughts.

**Using the GitHub Search API**
To get around this problem in `github-repo-automation`, I opted to use the GitHub Search API to identify the relevant repositories:
https://developer.github.com/v3/search/#search-repositories

The query we used there looks like this:
```
org:googleapis language:typescript language:javascript is:public archived:false
```

**Inversion of control**
Maybe a better question is, why do we need to have a centralized place for these jobs to run.  Should we consider pushing responsibility of registration down to the repository level?  I could absolutely see a future where orgs outside of googleapis (ads, analytics) want to use this tool. Is there a reason we couldn't ask repository owners to have a kokoro job on a cron that calls autosynth on their own?

SurferJeffAtGoogle: > I'd like to look at other ways to approach this problem. I am very worried about having yet another place where we have to list out all applicable targets. Two different thoughts.
> 
> **Using the GitHub Search API**
> To get around this problem in `github-repo-automation`, I opted to use the GitHub Search API to identify the relevant repositories:
> https://developer.github.com/v3/search/#search-repositories
> 
> The query we used there looks like this:
> 
> ```
> org:googleapis language:typescript language:javascript is:public archived:false
> ```

That's actually how I wrote the code the first time.  The problem is that all the queries came back marked incomplete, and indeed were missing a few results.  So autosynth would not regenerate a few unlucky repos every day.

https://developer.github.com/v3/search/#timeouts-and-incomplete-results

> **Inversion of control**
> Maybe a better question is, why do we need to have a centralized place for these jobs to run. Should we consider pushing responsibility of registration down to the repository level?

That's kinda what this code does for all but about 10 repos.  Register your new *nodejs* repo by adding synth.py to the root, and autosynth will start generating it.

If we pushed kokoro registration down to the repo levels, then we'd be maintaining 300+ kokoro job configs in piper, that have to stay in sync with 300+ build configs and synth.py files in the repos.

>I could absolutely see a future where orgs outside of googleapis (ads, analytics) want to use this tool. Is there a reason we couldn't ask repository owners to have a kokoro job on a cron that calls autosynth on their own?

I see no reason.  They can do that today.

JustinBeckwith: Ok - on the search result API thing, bummer, didn't know that, thank you :) I'm still wigged out about the 10 weird repos in the yaml.  Instead, could we use the repos list API, find all TypeScript/JavaScript APIs with a `synth.py`, and use those?
https://developer.github.com/v3/repos/#list-organization-repositories

I'm really worried we are going to create a new repo without the `nodejs-` prefix, try to start using synthtool, and forget to register it here.  Thinking out loud - I wonder if a model like this wouldn't be better to drive the entire autosynth process.  Grab every repo in the org, check if it has a synth.py, and run it.

autosynth/github.py
```diff
@@ -353,6 +369,47 @@ def update_pull_labels(
...
+    def list_repos(self, org: str) -> List[str]:
+        """Returns a list of all the repositories in an organization.
+
+        Args:
+            org (str): The name of the organization.
+
+        Returns:
+            List[str]: The list of repository names.
+        """
+        url = f"{_GITHUB_ROOT}/orgs/{org}/repos?type=public"
```
JustinBeckwith, line 381: You probably also want to filter out archived repositories

JustinBeckwith:
Haha, this will create a little trouble for Java:
https://github.com/googleapis?q=&type=&language=shell

There are a few repositories GitHub Linguist thinks are shell scripts due to kokoro being so chonky.  This is pretty easily fixed with a `.gitattributes` file filtering out the `.kokoro` directory:
https://github.com/github/linguist#overrides

SurferJeffAtGoogle:
> Haha, this will create a little trouble for Java:
> https://github.com/googleapis?q=&type=&language=shell
> 

I added code to only consider the 7 silver languages, and ignore languages like Shell.

Most of these repos will be correctly collected because they contain the language name in their names.  Inspecting the languages in the repo is a fall-back.

> There are a few repositories GitHub Linguist thinks are shell scripts due to kokoro being so chonky. This is pretty easily fixed with a `.gitattributes` file filtering out the `.kokoro` directory:
> https://github.com/github/linguist#overrides

autosynth/github.py
SurferJeffAtGoogle: Done.

PR #650, opened 2020-06-26
chore: use GOOGLE_CLOUD_PROJECT in samples noxfile

busunkim96:


synthtool/gcp/templates/python_samples/noxfile.py.j2
```diff
@@ -72,7 +72,7 @@ def get_pytest_env_vars():
     env_key = TEST_CONFIG['gcloud_project_env']
     # This should error out if not set.
     ret['GOOGLE_CLOUD_PROJECT'] = os.environ[env_key]
-    ret['GCLOUD_PROJECT'] = os.environ[env_key]
+    ret['GCLOUD_PROJECT'] = os.environ[env_key]  # deprecated
```
tmatsuo, line 75: I think you can also remove this line if you're sure it is not used.

PR #658, opened 2020-07-11
fix: update flake8 to match black

crwilcox:
Black disobeys E231 (adding comma to end, but no whitespace)


crwilcox:
Closing this. I think the culprit is actually the fixup script.

PR #661, opened 2020-07-14
fix: fix cov_level issue

arithmetic1728:
if `cov_level` is not provided `common.py_library`, then it will be an undefined variable in [cover section in noxfile.py.j2](https://github.com/googleapis/synthtool/blob/master/synthtool/gcp/templates/python_library/noxfile.py.j2#L164). Therefore, `--fail-under={{ cov_level if cov_level != None else '100' }}` will become `--fail-under=` instead of `--fail-under=100`.

The fix is to set `cov_level` to None if not provided.

### DIFF
--- a/synthtool/gcp/common.py
+++ b/synthtool/gcp/common.py
@@ -91,6 +91,10 @@ class CommonTemplates:
                     "system_test_python_versions"
                 ]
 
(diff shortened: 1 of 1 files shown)


arithmetic1728:
@busunkim96

PR #666, opened 2020-07-14
feat(java): allow generating non-cloud artifacts

chingor13:
This will allow us to start generating APIs like analytics.

Currently, we assume that all clients are cloud clients and include "cloud" in the artifact name. This change allows the synth.py to set `cloud_api=False` so that we don't prefix with `cloud-`.

It's possible to do this by using `destination_name`, however that would mean a breaking behavior change and we would need to update many `synth.py` files across the java repos.

The new call would look something like:
```python3
java.bazel_library(
  service=service,
  version=version,
  proto_path=f'google/analytics/admin/{version}',
  bazel_target=f'//google/analytics/admin/{version}:google-{service}-{version}-java',
  cloud_api=False,
)
```


SurferJeffAtGoogle:
In the PR description, could you describe how the code changes allow generating non-cloud artifacts?  I don't see the connection.

chingor13:
> In the PR description, could you describe how the code changes allow generating non-cloud artifacts? I don't see the connection.

Updated the description

PR #667, opened 2020-07-15
chore: add config files for cloud-rad for node.js

fhinkel:
Generate and upload yaml files for ref docs.

### DIFF
--- a/synthtool/gcp/templates/node_library/api-extractor.json
+++ b/synthtool/gcp/templates/node_library/api-extractor.json
@@ -1,4 +1,4 @@
-/**
+/***
  * Config file for API Extractor.  For more info, please visit: https://api-extractor.com
  */
 {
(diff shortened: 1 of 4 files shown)


synthtool/gcp/templates/node_library/api-extractor.json
```diff
@@ -0,0 +1,369 @@
+/**
+ * Config file for API Extractor.  For more info, please visit: https://api-extractor.com
+ */
+{
+  "$schema": "https://developer.microsoft.com/json-schemas/api-extractor/v7/api-extractor.schema.json",
+
+  /**
```
bcoe, line 7: should this be a `.js` file, seems to be very angry about the  comments in the JSON.

fhinkel: It's supposed to be a `json` file, generated by api-extactor. I added a `.gitattributes` file instead so it looks less angry.

PR #669, opened 2020-07-16
fix: re-enable obsolete tracking with a new strategy

SurferJeffAtGoogle:
old strategy: observe file timestamps to see which files were touched
new strategy: use watchdog to observe every file system event

Control tracking obsolete files with an environment variable.

Fixes https://github.com/googleapis/synthtool/issues/216

### DIFF
--- a/noxfile.py
+++ b/noxfile.py
@@ -37,6 +37,6 @@ def lint(session):
 
 @nox.session(python=['3.6', '3.8'])
 def test(session):
-    session.install('pytest', 'pytest-cov', 'requests_mock')
+    session.install('pytest', 'pytest-cov', 'requests_mock', 'watchdog')
     session.run('pip', 'install', '-e', '.')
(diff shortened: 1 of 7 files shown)


SurferJeffAtGoogle:
Here's the output from an example run:
https://github.com/googleapis/google-cloud-ruby/pull/7049/files#diff-6affd48be461b5ccaa75e24a0c8f4c5c

SurferJeffAtGoogle:
ping @chingor13

chingor13:
It's hard to tell by reading code, but this will be able to catch files that are generated but not updated (because the file date has been touched, but `git diff` won't see the change)?

SurferJeffAtGoogle:
> It's hard to tell by reading code, but this will be able to catch files that are generated but not updated (because the file date has been touched, but `git diff` won't see the change)?

If my understanding is correct, it will even catch files that are generated and the date has been left unchanged, which happens when you use the OS's copy command.

PR #676, opened 2020-07-17
Dependencies Check Update

salehsquared:
Changes the dependencies.sh kokoro script to only compare 'compile' and 'runtime' scoped dependencies between the flattened POM and original POM. 

This follows an issue recently noted within the java-bigquerystorage libraries in which inconsistent behavior with 'provided' scope transitive dependencies.


salehsquared:
@googlebot I signed it!

PR #678, opened 2020-07-20
build: update dependencies check to only check for runtime and compile scopes

salehsquared:
Changes the dependencies.sh kokoro script to only compare 'compile' and 'runtime' scoped dependencies between the flattened POM and original POM.

This follows an issue recently noted within the java-bigquerystorage libraries in which inconsistent behavior with 'provided' scope transitive dependencies.


stephaniewang526:
this should be renamed to be `feat: update dependencies check to only check for runtime and compile scopes`

stephaniewang526:
@saturnism @elharo this is for verifying flattened pom vs. original pom completeness

synthtool/gcp/templates/java_library/.kokoro/dependencies.sh
```diff
@@ -43,12 +43,13 @@ function completenessCheck() {
...
+  # Only dependencies with 'compile' or 'runtime' scope are included from original dependency list.
   msg "Generating dependency list using original pom..."
-  mvn dependency:list -f pom.xml -Dsort=true | grep '\[INFO]    .*:.*:.*:.*:.*' | sed -e s/\\s--\\smodule.*// | grep -v ':test$' >.org-list.txt
+  mvn dependency:list -f pom.xml -DincludeScope=runtime -Dsort=true | grep '\[INFO]    .*:.*:.*:.*:.*' | sed -e s/\\s--\\smodule.*// >.org-list.txt
 
-  # Output dep list generated using the flattened pom (test scope deps are ommitted)
+  # Output dep list generated using the flattened pom (only 'compile' and 'runtime' scopes)
   msg "Generating dependency list using flattened pom..."
-  mvn dependency:list -f .flattened-pom.xml -Dsort=true | grep '\[INFO]    .*:.*:.*:.*:.*' >.new-list.txt
+  mvn dependency:list -f .flattened-pom.xml -DincludeScope=runtime -Dsort=true | grep '\[INFO]    .*:.*:.*:.*:.*' >.new-list.txt
```
saturnism, line 52: for the flatten one, should we not limit the scope? I.e., flattened pom should probably only have runtime artifacts even w/o scoping.

elharo: compile time is needed too

elharo: compile is needed too

stephaniewang526: Based on the [doc](https://maven.apache.org/plugins/maven-dependency-plugin/list-mojo.html#includeScope):

runtime scope gives runtime and compile dependencies

stephaniewang526: > for the flatten one, should we not limit the scope? I.e., flattened pom should probably only have runtime artifacts even w/o scoping.

This depends on what we do in the flatten plugin -- if we keep direct provides deps, flattened pom will have them. I think it is safer for us to do a strict comparison btw the original pom and the flattened pom here.

salehsquared: This was my concern as well - I think making this change to the PR would depend on the status of the provided transitive PR on the flatten-maven-plugin, since we don't certainly know yet if direct provided dependencies will be included.

stephaniewang526:
@chingor13 could you merge this one please?

PR #680, opened 2020-07-21
fix: move gitattributes file to nodejs templates

fhinkel:


.gitattributes
```diff
@@ -1,2 +0,0 @@
-**/api-extractor.json  linguist-language=JSON-with-Comments
```
SurferJeffAtGoogle, line 1: Because this repo still contains an `api-extractor.json`, I see a benefit in preserving this .gitattributes.

fhinkel: @SurferJeffAtGoogle The rule was moved to a subfolder so it works both for this repo and the target repos. Looking at `api-extractor.json` on this branch, it looks OK: https://github.com/googleapis/synthtool/blob/moveConfigFile/synthtool/gcp/templates/node_library/api-extractor.json

PR #657, opened 2020-07-09
feat(python_samples): Added functionality & updated py sample template

runargs:
Updated template to be used for Python sample libraries, as well as adding tests and functionality to generate the template which will be utilized in the future but do not currently affect documentation generation.

PR #690, opened 2020-07-23
Revert "feat(python_samples): Added functionality & updated py sample template"

busunkim96:
Reverts googleapis/synthtool#657

This is currently breaking generations for libraries that add samples. It also results in two READMEs (README.md and README.rst) in the same directory.


kolea2:
@busunkim96 would my fix resolve this? https://github.com/googleapis/synthtool/pull/685 or is there another case I hadn't considered?

busunkim96:
@kolea2 Ah yep, that should work!

PR #693, opened 2020-07-24
docs: recommend installation via pip -e

### LINKED ISSUE #691: Lost installation documentation
When merging with autosynth and shuffling the docs around, it seems we lost the documentation on how to install from head.

busunkim96:
Closes #691

### DIFF
--- a/synthtool/README.md
+++ b/synthtool/README.md
@@ -16,11 +16,11 @@ This tool helps to generate and layout cloud client libraries. Synthtool runs th
 
 3. **Docker**  Some synth.py files require [Docker] to generate code.
 
-4. Install this library with pip:
+4. Clone this repository and install this library with pip:
 
     ```
     cd synthtool
-    export PYTHONPATH=`pwd`
+    python3 -m pip install -e .
     ```
 
 


synthtool/README.md
```diff
@@ -16,11 +16,11 @@ This tool helps to generate and layout cloud client libraries. Synthtool runs th
 
 3. **Docker**  Some synth.py files require [Docker] to generate code.
 
-4. Install this library with pip:
+4. Clone this repository and install this library with pip:
 
     ```
     cd synthtool
-    export PYTHONPATH=`pwd`
+    python3 -m pip install -e .
```
SurferJeffAtGoogle, line 23: Earlier versions of this file had the exact same command, but the command failed for Justin Beckwith.

So line 19 looks good, but line 23 doesn't work for all users.

busunkim96: @JustinBeckwith Do you remember how this failed for you?

JustinBeckwith: Oh gosh, this was a while ago now, and I can't remember what was failing.  What I *do* remember is that when installing directly from the repository, `synthtool --version` became practically useless.  I was struggling to get the right version of `synthtool` on my path, and doing a `--version` would always tell me the same version regardless of which commit I had installed against.  

I'd like to arrive at one of two places here. Either:
- We update `synthtool --version` to include the sha at which it's running alongside the current version or
- We arrive at a place where we're consistently releasing every time after updates are made

SurferJeffAtGoogle:
Actually, Justin's difficulty was running autosynth, not synthtool, so I'm ok with this change.

PR #695, opened 2020-07-24
feat(python_samples): Custom path for generation

runargs:
Added functionality to be used for generating documentation for Python samples. A demonstration of the documentation this code can render (Using the README.md template under [/python_samples/](https://github.com/googleapis/synthtool/tree/master/synthtool/gcp/templates/python_samples)) is available [here](https://github.com/runargs/python-docs-samples/tree/synth-DEMO/bigtable).

This function will default to generating the README in the current directory if the synth.py file is put in a sample folder, and or the "/samples" folder if it is placed in the top level of a client library repo. A custom path for where to generate may also be specified through the "sample_project_dir" key in .repo.metadata.json


synthtool/gcp/common.py
```diff
@@ -59,11 +59,37 @@ def _generic_library(self, directory: str, **kwargs) -> Path:
...
+        # load common repo meta information (metadata that's not language specific).
+        if "metadata" in kwargs:
+            self._load_generic_metadata(kwargs["metadata"])
+        in_client_library = Path("samples").exists()
+        if sample_project_dir == ".": #if set to default value
+            if in_client_library:
+                sample_project_dir = "samples"
+            elif "sample_project_dir" in kwargs["metadata"]["repo"]:
+            # a custom path may be specified
+                if Path(sample_project_dir).exists():
```
kolea2, line 85: what happens if the path does not exist?

synthtool/gcp/common.py
```diff
@@ -59,11 +59,37 @@ def _generic_library(self, directory: str, **kwargs) -> Path:
...
+        """
         # kwargs["metadata"] is required to load values from .repo-metadata.json
         if "metadata" not in kwargs:
             kwargs["metadata"] = {}
-        return self._generic_library("python_samples", **kwargs)
+        # load common repo meta information (metadata that's not language specific).
+        if "metadata" in kwargs:
+            self._load_generic_metadata(kwargs["metadata"])
+        in_client_library = Path("samples").exists()
+        if sample_project_dir == ".": #if set to default value
```
kolea2, line 80: why do we need this check? Isn't that the default value?

synthtool/gcp/common.py
```diff
@@ -59,11 +59,37 @@ def _generic_library(self, directory: str, **kwargs) -> Path:
...
+        Args:
+            sample_project_dir (Union[Path, str]): The directory for samples, where
+            the README is to be generated
+        """
         # kwargs["metadata"] is required to load values from .repo-metadata.json
         if "metadata" not in kwargs:
             kwargs["metadata"] = {}
-        return self._generic_library("python_samples", **kwargs)
+        # load common repo meta information (metadata that's not language specific).
+        if "metadata" in kwargs:
```
kolea2, line 77: can we remove this if statement, since we just set metadata on line 75?

synthtool/gcp/common.py
```diff
@@ -59,11 +59,37 @@ def _generic_library(self, directory: str, **kwargs) -> Path:
 
         return result
 
-    def py_samples(self, **kwargs) -> Path:
+    def py_samples(self, sample_project_dir: PathOrStr = ".", **kwargs) -> Path:
```
kolea2, line 62: can we add some unit tests here?

synthtool/gcp/common.py
runargs: It just goes with the default "."

synthtool/gcp/common.py
runargs: I added this for handling the case sample_proj_dir is set by parameter as opposed to through the metadata. I'm open to removing it if that doesn't seem necessary though.

synthtool/gcp/common.py
kolea2: Hmm, should it throw an exception if the user specifies a directory that is not present? Otherwise it may be some unexpected behavior to the user who is expecting a readme in the specified directory.

synthtool/gcp/common.py
kolea2: Gotcha, makes sense. I think a unit test illustrating the user setting the parameter for the directory would be a good addition. In addition - a test with both the parameter set and repo metadata set would be good (to show which takes precedence).

synthtool/gcp/common.py
```diff
@@ -60,10 +60,36 @@ def _generic_library(self, directory: str, **kwargs) -> Path:
...
         # kwargs["metadata"] is required to load values from .repo-metadata.json
         if "metadata" not in kwargs:
             kwargs["metadata"] = {}
-        return self._generic_library("python_samples", **kwargs)
+        # load common repo meta information (metadata that's not language specific).
+        self._load_generic_metadata(kwargs["metadata"])
+
+        in_client_library = Path("samples").exists()
+        sample_project_dir = kwargs.get('metadata').get('repo').get('sample_project_dir')
+        #None if custom path not specified
```
kolea2, line 77: I'd move this above the line of code instead of after

.gitignore
```diff
@@ -114,3 +114,8 @@ venv.bak/
 # synthtool artifacts
 working_repo/
 sponge_log.xml
+
+# synthtool testing artifacts
+repo.metadata.json
```
kolea2, line 119: double check repo metatata and synth.py

synthtool/gcp/common.py
```diff
@@ -60,10 +61,34 @@ def _generic_library(self, directory: str, **kwargs) -> Path:
...
+        if sample_project_dir is None: #Not found in metadata
+            if in_client_library:
+                sample_project_dir = "samples"
+            else:
+                sample_project_dir = "."
+        elif not Path(sample_project_dir).exists():
+            raise Exception(f"'{sample_project_dir}' does not exist")
+         
+        logger.debug(f"Generating templates for samples directory '{sample_project_dir}'")
+        py_samples_templates = Path(self._template_root) / "python_samples" / "new_README"
```
kolea2, line 87: let's look into why this changed

tests/test_common.py
```diff
@@ -29,3 +33,30 @@ def test_handles_acronym():
 def test_handles_empty_string():
     assert decamelize(None) == ""
     assert decamelize("") == ""
+
+def test_py_samples_clientlib():
+    path_to_gen = MOCK / "client_library"
+    os.chdir(path_to_gen)
+    os.system("python3 -m synthtool")
```
kolea2, line 40: take a look and see if this can be replaced with directly calling `py_samples`: https://github.com/googleapis/synthtool/blob/master/tests/test_templates.py#L111-L123

synthtool/gcp/common.py
```diff
@@ -18,13 +18,14 @@
 import yaml
 from pathlib import Path
 from typing import Dict, List, Optional
+from sys import exit
```
kolea2, line 21: remove

kolea2 (changes requested):

kolea2 (approved):
LGTM. @SurferJeffAtGoogle can you take a look as well?

busunkim96 (approved):

PR #701, opened 2020-07-29
docs: add links to the CHANGELOG from the README.md for Java and Node

feywind:
This is for making this user-requested change global across repos that have a README.

https://github.com/googleapis/nodejs-pubsub/pull/1067

I didn't see equivalents for the other languages... is that something we want to figure out? Or just go with these two for now?


feywind:
Ahh the joys of learning a new keyboard :D

feywind:
I'm moving this back to just Node, because that's the immediate need, and most of the Java repos don't seem to use the generated README anyway.

synthtool/gcp/templates/node_library/README.md
```diff
@@ -24,6 +24,9 @@
 {{ metadata['description'] }}
 {% endif %}
 
+A fairly comprehensive list of changes in each version may be found in
```
bcoe, line 27: nit:, I might just be definitive, and say "A comprehensive list of changes".

PR #700, opened 2020-07-29
feat(python-library): changes to docs job

tmatsuo:
* migrate to Trampoline V2
* add docs-presubmit job
* create docfx yaml files and upload them to another bucket

PR #707, opened 2020-07-30
feat(python-library): changes to docs job

tmatsuo:
This is another trial of #700 

- [x] add a test that fails
- [x] make the test pass

### DIFF
--- a/synthtool/gcp/common.py
+++ b/synthtool/gcp/common.py
@@ -105,6 +105,7 @@ class CommonTemplates:
         if "samples" not in kwargs:
             self.excludes += ["samples/AUTHORING_GUIDE.md", "samples/CONTRIBUTING.md"]
 
+        self.excludes += [".kokoro/docker/docs/fetch_gpg_keys.sh"]
         return self._generic_library("python_library", **kwargs)
 
     def java_library(self, **kwargs) -> Path:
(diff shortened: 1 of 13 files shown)


tmatsuo:
[The build](https://source.cloud.google.com/results/invocations/3effcd5e-eced-4719-9ff2-f871f788d2c2/targets) failed as expected:

```
E   jinja2.exceptions.TemplateSyntaxError: Missing end of comment tag
```

synthtool/gcp/common.py
```diff
@@ -105,6 +105,7 @@ def py_library(self, **kwargs) -> Path:
         if "samples" not in kwargs:
             self.excludes += ["samples/AUTHORING_GUIDE.md", "samples/CONTRIBUTING.md"]
 
+        self.excludes += [".kokoro/docker/docs/fetch_gpg_keys.sh"]
```
busunkim96, line 108: If you add it to the excludes list the file will not be copied over. I think escaping the comment string should work? https://jinja.palletsprojects.com/en/2.11.x/templates/#escaping

tmatsuo: Ya, I used `raw`. It looks fine.

busunkim96 (approved):

PR #709, opened 2020-07-30
feat: when obsolete file tracking gets flipped on, and there are no c…

SurferJeffAtGoogle:
…hanges to the generated code, create a pull request anyway

so that future files that get deleted from templates will be properly removed from target repos

### DIFF
--- a/autosynth/synth.py
+++ b/autosynth/synth.py
@@ -24,7 +24,7 @@ import subprocess
 import sys
 import tempfile
 import typing
-from typing import Dict, Sequence
+from typing import Dict, Optional, Sequence
 
 import synthtool.sources.git as synthtool_git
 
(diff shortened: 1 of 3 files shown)


autosynth/synth.py
```diff
@@ -357,6 +370,26 @@ def synthesize_loop(
...
+    # Synthesize the library with the most recent versions of all sources.
+    youngest = len(toolbox.versions) - 1
+    has_changes = toolbox.synthesize_version_in_new_branch(synthesizer, youngest)
+    if not has_changes:
+        if not toolbox.metadata_contains_generated_files(
+            toolbox.branch
+        ) and toolbox.metadata_contains_generated_files(toolbox.sub_branch(youngest)):
+            # Special case: the repo owner turned on obsolete file tracking.
+            # Generate a one-time PR containing only metadata changes.
+            executor.check_call(["git", "checkout", toolbox.branch])
```
bcoe, line 382: @SurferJeffAtGoogle if we turn this setting on globally, I gather we'll get a one time PR across all our repositories?

PR #710, opened 2020-07-31
fix(python-library): add missing changes

tmatsuo:
Somehow some changes are missing from the template.

### DIFF
--- a/synthtool/gcp/templates/python_library/.kokoro/publish-docs.sh
+++ b/synthtool/gcp/templates/python_library/.kokoro/publish-docs.sh
@@ -18,26 +18,16 @@ set -eo pipefail
 # Disable buffering, so that the logs stream through.
 export PYTHONUNBUFFERED=1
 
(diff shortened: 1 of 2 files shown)


synthtool/gcp/templates/python_library/.kokoro/publish-docs.sh
```diff
@@ -18,26 +18,16 @@ set -eo pipefail
 # Disable buffering, so that the logs stream through.
 export PYTHONUNBUFFERED=1
 
-cd github/{{ metadata['repo']['repo'].split('/')[1] }}
```
busunkim96, line 21: Does trampoline handle `cd`ing into the directory?

tmatsuo: yes

tmatsuo:
@busunkim96 FYI, I run synthtool manually with texttospeech and got the expected result:
https://github.com/googleapis/python-texttospeech/pull/73

The docs job failed, but it failed for different reason.

PR #714, opened 2020-08-06
fix: don't remove obsolete file if it matches excludes pattern

SurferJeffAtGoogle:


synthtool/metadata.py
```diff
@@ -56,6 +62,13 @@ def add_git_source(**kwargs) -> None:
     _metadata.sources.add(git=metadata_pb2.GitSource(**kwargs))
 
 
+def add_pattern_excluded_during_copy(file_path: str) -> None:
```
chingor13, line 65: Can we rename this to `glob_pattern` or indicate that this can be a `glob`?

PR #715, opened 2020-08-06
fix: don't remove obsolete files when synth raises exception

SurferJeffAtGoogle:


tests/test_metadata.py
```diff
@@ -216,6 +216,28 @@ def test_nothing_happens_when_disabled(source_tree, preserve_track_obsolete_file
     assert os.path.exists("code/c")
 
 
+def test_nothing_happens_when_exception(source_tree, preserve_track_obsolete_file_flag):
```
bcoe, line 219: in the case of `@google-cloud/logging` what state would this leave us in once the run finishes, will we move  towards  being in a health up-to-date state with the metadata?

SurferJeffAtGoogle: It will leave the old synth.metadata without updating it at all.

PR #717, opened 2020-08-06
build(test): recursively find test files; fail on unsupported dependency versions

feywind:
... it seems to have been the case in the past.

We discovered that a big swath of tests weren't being run in the nodejs-pubsub library because they're in a subtree. This would enable recursion explicitly in the Mocha config file. I have to believe that we once did this, because the tests were there.


fhinkel (approved):

bcoe:
Having a discussion with @feywind about perhaps also updating our `npm i` command when we land this, such that it captures dependency issues.

bcoe (approved):

PR #724, opened 2020-08-14
feat(python_samples): Allow for additional README generation in specified path(s)

runargs:
Added functionality to be used for generating documentation for Python samples.

This PR introduces a feature such that users can now define "override" folders in any sample's metadata.
For example, if one wanted to generate a README under the typical-case "/samples" directory but **also** "/samples/snippets" and "/samples/quickstart", that is now possible.

READMEs generated will only include the information pertaining to the specific samples that have the override path defined, such that "/samples/quickstart" for example would only have README information about the Quickstart, and not any other sample.

Also includes tests/mock files


synthtool/gcp/common.py
```diff
@@ -62,18 +64,23 @@ def _generic_library(self, directory: str, **kwargs) -> Path:
...
-        'samples' folder. A custom path for where to generate may also be specified.
-        Renders README.md according to .repo-metadata.json
+        Handles generation of README.md templates for Python samples
+        - Determines whether generation is being done in a client library or in a samples
+        folder automatically
+        - Otherwise accepts manually set sample_project_dir through kwargs metadata
+        - Delegates generation of additional sample documents alternate/overridden folders
+        through py_samples_override()
         """
+        logger.debug(f"start")
```
kolea2, line 76: don't think we need this anymore

synthtool/gcp/common.py
```diff
@@ -97,15 +105,77 @@ def py_samples(self, **kwargs) -> Path:
...
+                cur_override_sample = override_paths_to_samples.get(override_path)
+                # Base case: No samples are yet planned to gen in this override dir
+                if cur_override_sample == None:
+                    override_paths_to_samples[override_path] = [sample]
+                # Else: Sample docs will be generated in README merged with other
+                # sample doc(s) already planned to generate in this dir
+                else:
+                    cur_override_sample.append(sample)
+                    override_paths_to_samples[override_path] = cur_override_sample
+            else:
```
kolea2, line 126: maybe add a comment here?

synthtool/gcp/common.py
```diff
@@ -97,15 +105,77 @@ def py_samples(self, **kwargs) -> Path:
...
+                default_samples_dict.append(sample)
+
+        result = (
+            []
+        )  # List of paths to tempdirs which will be copied into sample folders
+        copy_kwargs = deepcopy(
```
kolea2, line 132: maybe rename `copy_kwargs` to something like `overridden_samples_kwargs`?

runargs: Good idea! more descriptive, will do :)

tests/generationmock/multiple_override_path/.repo-metadata.json
```diff
@@ -0,0 +1,33 @@
...
+    "client_documentation": "https://googleapis.dev/python/bigtable/latest",
+    "issue_tracker": "https://issuetracker.google.com/savedsearches/559777",
+    "release_level": "ga",
+    "language": "python",
+    "repo": "googleapis/python-bigtable",
+    "distribution_name": "google-cloud-bigtable",
+    "api_id": "bigtable.googleapis.com",
+    "requires_billing": true,
+    "client_library": true,
+    "samples": [
```
bcoe, line 14: we did this slightly differently with Node.js, where we allowed for meta information to  be provided in the prefix of the sample, see:

https://github.com/googleapis/nodejs-storage/blob/master/samples/bucketMetadata.js#L17

Does  this serve a similar purpose to these comments?

```js
// sample-metadata:
//   title: Storage Get Bucket Metadata.
//   description: Get bucket metadata.
//   usage: node bucketMetadata.js <BUCKET_NAME>
```

runargs: Yes, it does! It looks about the same except this uses Synthtool which utilizes the .repo-metadata.json files. I'm not familiar with Node's libraries though, do you know what tool is being used for this generation? Or is it just from a different section of synthtool?

bcoe: This is handled with `synthtool` too, you can find the PR where I added this [here](https://github.com/googleapis/synthtool/pull/230).

I  was thinking it allows  you to keep this information close to the sample itself. If you like this idea, it might be worth considering this approach. 

If you think there's merit to having both approaches,  I totally understand too 👍 

tldr; whatever can best help serve your work on the Python samples.

PR #732, opened 2020-08-20
fix(python_samples): README link fix, enforce samples=True 

runargs:
Updated sources link to direct to specified product's repo, enforced `kwargs["metadata"]["samples"] = True` every time `py_samples()` is run, as it's a given if the function is called and shouldn't have to be specified from the user side.
Oh also the wording was slightly redundant.
TLDR, Just some wording fixes and a one line edit

### DIFF
--- a/synthtool/gcp/common.py
+++ b/synthtool/gcp/common.py
@@ -90,7 +90,7 @@ class CommonTemplates:
             ]
         )
 
-        # enit sure samples will generate
+        # ensure samples will generate
         kwargs["metadata"]["samples"] = True
 
         # determine if in client lib and set custom root sample dir if specified, else None
--- a/synthtool/gcp/common.py
+++ b/synthtool/gcp/common.py
@@ -17,9 +17,9 @@ import os
 import re
 import shutil
 import yaml
+from copy import deepcopy
 from pathlib import Path
 from typing import Dict, List, Optional
-
 import jinja2
 
 from synthtool import _tracked_paths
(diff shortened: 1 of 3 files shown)


kolea2 (approved):

synthtool/gcp/common.py
```diff
@@ -90,6 +90,9 @@ def py_samples(self, **kwargs) -> List[Path]:
             ]
         )
 
+        # ensure samples will generate
+        kwargs["metadata"]["samples"] = True
```
kolea2, line 94: what happens if this is not set?

synthtool/gcp/templates/python_samples/README.md
```diff
@@ -37,15 +37,14 @@ To run this sample:
 {% if metadata['repo']|length %}{% if metadata['repo']['client_library'] %}
 These samples use the [Google Cloud Client Library for Python][client_library_python].{% endif %}
 You can read the documentation for more details on API usage and use GitHub
-to [browse the source][source] and [report issues][issues].{% endif %}
+to <a href="https://github.com/{{ metadata['repo']['repo'] }}">browse the source</a> and [report issues][issues].{% endif %}
```
kolea2, line 40: can we add a quick test that the url is correct?

kolea2 (changes requested):

synthtool/gcp/common.py
runargs: if it isn't set, _generic_library() will add README.md to the `excludes` https://github.com/googleapis/synthtool/blob/ca275fc17e8e92c81c04f19de1d22e016be3f0e2/synthtool/gcp/common.py#L57

synthtool/gcp/templates/python_samples/README.md
```diff
@@ -37,15 +37,14 @@ To run this sample:
...
 
 ### Contributing
-For [contributing guidelines][contrib_guide], the [Python style guide][py_style], and more information on prerequisite steps to contribute, view the source code at <a href="https://github.com/{{ metadata['repo']['repo'] }}">{{ metadata['repo']['repo'] }}</a>.
+View the [contributing guidelines][contrib_guide], the [Python style guide][py_style]. More information on prerequisite steps to contribute may be available at the source code.
```
kolea2, line 43: I would just make this read, "View the contributing guidelines and the Python style guide for more information."

kolea2 (changes requested):

kolea2 (approved):

PR #742, opened 2020-08-27
chore(py-library): enable snippet-bot

tmatsuo:


SurferJeffAtGoogle:
Before we enable it for all Python libraries, have we tried enabling it for a single library?

tmatsuo:
@SurferJeffAtGoogle Yes, 
https://github.com/googleapis/python-texttospeech
https://github.com/googleapis/python-vision

The bot works for both.

Also, the bot is very quiet and harmless, so I'm not concerned about rolling back.

tmatsuo:
@SurferJeffAtGoogle Can we merge?

PR #743, opened 2020-08-27
chore: enable obsolete file tracking for php

SurferJeffAtGoogle:


SurferJeffAtGoogle:
I discussed with David Supplee.  He has reason to believe that protoc may stop generating some files it should continue generating.  Enabling obsolete file tracking in that scenario would break users.  So, we'll leave this feature disabled.

PR #745, opened 2020-08-27
chore(py-library): update decrypt secrets file

busunkim96:
From https://github.com/GoogleCloudPlatform/python-docs-samples/blob/master/scripts/decrypt-secrets.sh


synthtool/gcp/templates/python_library/scripts/decrypt-secrets.sh
```diff
@@ -20,14 +20,25 @@ ROOT=$( dirname "$DIR" )
 # Work from the project root.
 cd $ROOT
 
+# Prevent it from overriding files.
```
SurferJeffAtGoogle, line 23: Please expand the comment.
What is 'it'?
Why do we want to prevent it from overriding (overwriting?) files?

tmatsuo: I think it's copied from python-docs-samples.

Here is some background.

We recommend that sample authors use their own service account files and cloud project. In that case, they are supposed to prepare these files by themselves.

I don't want `decrypt-secrets.sh` blindly overwrite those files.

SurferJeffAtGoogle: I agree that we should not overwrite those files.  Please copy your two-sentence background description into a comment.  In other words:

```py
# Prevent it from overriding files.
# We recommend that sample authors use their own service account files and cloud project.
# In that case, they are supposed to prepare these files by themselves.
```

busunkim96: Updated

SurferJeffAtGoogle (approved):

PR #749, opened 2020-09-01
fix: set SETUPTOOLS_USE_DISTUTILS

busunkim96:


SurferJeffAtGoogle:
Actually, I'm skeptical this will fix the issue.  When bazel kicks off, I think it ignores all environment variables. Do @alexander-fenster  and @vam-google have any insight?

vam-google:
Yes, it should be ignored. There is [--action_env](https://docs.bazel.build/versions/master/command-line-reference.html#flag--action_env) command line flag to specify environment variable for bazel actions (append to bazel build, can be added to the .bazelrc ), but I have never tried and not sure if that will work for repository rule (where it fails now), but it is worth trying.

PR #765, opened 2020-09-21
chore(python): add sphinx doctest extension

busunkim96:
Some libraries use this sphinx extension.


tseaver (approved):
FBO at least `python-datastore`

tmatsuo (approved):

PR #771, opened 2020-09-22
chore(python): skip reporting coverage for namespace package

tseaver:


busunkim96 (approved):
If it is a microgenerator library, the .coveragerc may be coming from https://github.com/googleapis/gapic-generator-python/blob/master/gapic/templates/.coveragerc.j2

PR #788, opened 2020-10-07
fix(java-templates): custom templates needs to load partial metadata

### LINKED ISSUE #791: Split Java README job is not loading readme-partials.yaml
Example: https://github.com/googleapis/java-spanner/pull/489

We are using `CommonTemplates.render()` which does not load the readme-partials metadata.

chingor13:
Refactors partial parsing into a separate module for re-use.

For java, we are using `CommonTemplates.render()` which does not include any magic metadata loading like `CommonTemplates._generic_library()` does. Instead allow the partial yaml parsing to be used externally and explicitly load the partial metadata in the `java.custom_templates()` helper. This fixes the README generation from our README autosynth.

Fixes #791


thiagotnunes (approved):

stephaniewang526 (approved):

bcoe (approved):

synthtool/gcp/partials.py
```diff
@@ -0,0 +1,44 @@
...
+from typing import Dict, List
+
+# these are the default locations to look up
+_DEFAULT_PARTIAL_FILES = [".readme-partials.yml", ".readme-partials.yaml"]
+
+
+def load_partials(files: List[str] = _DEFAULT_PARTIAL_FILES) -> Dict:
+    """
+    hand-crafted artisinal markdown can be provided in a .readme-partials.yml.
+    The following fields are currently supported:
```
bcoe, line 27: I like this refactor 👍 

It might be worth running the branch against one Node.js repo that uses partials, just to make sure it behaves as expected.

chingor13: Ran this against nodejs-storage and it succeeded with no changes to the README.md

PR #790, opened 2020-10-08
chore: clean up Node.js TOC for cloud-rad

fhinkel:


tbpg (approved):
I wonder how far we'll get with `sed` and not having to do full-on [customization](https://github.com/microsoft/rushstack/blob/master/apps/api-documenter/src/documenters/YamlDocumenter.ts).

SurferJeffAtGoogle (approved):

bcoe (approved):

PR #801, opened 2020-10-12
fix: load flags from the specified synth.py file

### LINKED ISSUE #793: autosynth: not respecting --pr-title flag
Java README generation is using a separate job that runs [`python3 -m autosynth.synth --pr-title="chore: regenerate README" <other args>`](https://github.com/googleapis/java-document-ai/blob/master/.kokoro/readme.sh). The first PR was correct, but [subsequent PRs](https://github.com/googleapis/java-document-ai/pull/209) have "[CHANGE ME] Re-generated to pick up changes in the API or client library generator." for the title.

We are not setting any flags, so in theory we should be hitting [this code path](https://github.com/googleapis/synthtool/blob/b6164c26a111f7f587099d31253abb96b5737bb2/autosynth/synth.py#L631-L651) that uses the pr title from the command line.

chingor13:
Fixes #793 

When we specify a custom synth.py file, we should parse the AST from that file when determining which flags to set. Note that the parse_flags method already supported picking a file, we just need to use the user input.

### DIFF
--- a/autosynth/synth.py
+++ b/autosynth/synth.py
@@ -617,7 +617,7 @@ def _inner_main(temp_dir: str) -> int:
 
         metadata_path = os.path.join(args.metadata_path or "", "synth.metadata")
 
-        flags = autosynth.flags.parse_flags()
+        flags = autosynth.flags.parse_flags(synth_file_name)
         # Override flags specified in synth.py with flags specified in environment vars.
         for key in flags.keys():
             env_value = os.environ.get(key, "")


bcoe (approved):

bcoe:
@chingor13 and I had an offline discussion about this PR; it unblocks quite a few Java PRs, so we're going ahead and landing, we will keep close eye on both Java and Node.js autosynth jobs tonight.

PR #803, opened 2020-10-13
fix: samples tests should pass if no samples exist

Daniel-Sanche:
Currently, `test-samples.sh` runs into a bug and crashes if the `./samples` directory doesn't exist in a repo. I'd argue that the check should pass automatically if there are no samples to test against. 

This would fix the issue we're running into on [python-logging](https://github.com/googleapis/python-logging), where no CI tests will pass until we merge the [PR adding samples](https://github.com/googleapis/python-logging/pull/67)

If we'd prefer having no samples to count as a test failure, I'd suggest we least add a better error message. Right now the code just crashes by trying to cd into a directory that doesn't exist

### DIFF
--- a/synthtool/gcp/templates/python_library/.kokoro/test-samples.sh
+++ b/synthtool/gcp/templates/python_library/.kokoro/test-samples.sh
@@ -66,38 +66,41 @@ set +e
 # Use RTN to return a non-zero value if the test fails.
 RTN=0
 ROOT=$(pwd)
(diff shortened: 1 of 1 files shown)


SurferJeffAtGoogle (approved):
LGTM.  Please wait for @busunkim96's approval.

busunkim96 (approved):
Thanks for adding this @Daniel-Sanche!

PR #805, opened 2020-10-14
refactor: move some code from synth.py to synth_toolbox.py

SurferJeffAtGoogle:
First step towards improved solution for coping with changes to target repos.

Also, lint tool updated many files.


bcoe (commented):
I'm guessing this work is in preparation to have one job that runs when code is merged to the main branch, and one job that runs nightly? am I following the motivation of the refactor?

SurferJeffAtGoogle:
> I'm guessing this work is in preparation to have one job that runs when code is merged to the main branch, and one job that runs nightly? am I following the motivation of the refactor?

Yes.  Both jobs will call the code in the new toolbox, but will share very little code left in synth.py.

autosynth/synth_toolbox.py
```diff
@@ -0,0 +1,379 @@
+#!/usr/bin/env python3.6
```
chingor13, line 1: Do you need this line? This file doesn't look executable.

chingor13 (approved):

PR #810, opened 2020-10-15
build: only check --engine-strict for production deps

bcoe:
Perform an installation of our `--production` dependencies first, with `--engine-strict`, so that we can detect dependencies that have dropped support for our supported Node.js version.


alexander-fenster (approved):

JustinBeckwith (approved):

chingor13 (approved):

bcoe:
going to test a little bit upstream before we merge.

bcoe:
seems to be working well, and will unblock some repos.

PR #819, opened 2020-10-19
fix(java_library): Update publish_javadoc.sh

lesv:
Upload from `target/devsite/reference` not `target/devsite`.

@tbpg FYI


chingor13 (approved):
Please make sure this merges as a "chore". Fix will tell the client library to ship a new patch version

lesv:
@chingor13 Sigh - just saw your comment about chore vs. fix - sigh / sorry.

Issue #812, opened 2020-10-16
introduce startTrackingObsoleteFiles()/stopTrackingObsoleteFiles()


bcoe:
We have been bumping into issues with tools like linters, where they might touch files that are outside of the responsibility of the code generator.

This teaches synthtool to track files that are not actually its responsibility. This in turn makes it difficult to implement the deletion logic.

## Proposal

Introduce `startTrackingObsoleteFiles()` and `stopTrackingObsoleteFiles()` which you use to "stop tracking obsolete files" before an operation like `lint --fix`.

```python
s.startTrackingObsoleteFiles()
// run the generator.
// copy templates.
s.stopTrackingObsoleteFiles()
// run post-processing like linting.
```

bcoe:
CC: @alexander-fenster, who I've talked to about this one.

SurferJeffAtGoogle:
Obsolete.

PR #824, opened 2020-10-21
feat: add start/stop_tracking_generated_files()

SurferJeffAtGoogle:
fixes https://github.com/googleapis/synthtool/issues/812


synthtool/gcp/common.py
```diff
@@ -269,6 +270,7 @@ def java_library(self, **kwargs) -> Path:
 
     def node_library(self, **kwargs) -> Path:
         # TODO: once we've migrated all Node.js repos to either having
+        metadata.start_tracking_generated_files()
```
bcoe, line 273: ~Calling `start_tracking_generated_files()` twice won't cause any issues? (_we will call it once for each API version._)~

_question answered below._

synthtool/metadata.py
```diff
@@ -224,38 +224,110 @@ def get_touched_file_paths(self) -> List[str]:
...
+
+    def clear_tracking_history(self):
+        """Clears the tracking history of generated files."""
+        self.generated_file_paths = []
+
+    def start_tracking(self):
+        """Create a watchdog observer and start tracking file changes."""
+        observer = getattr(self, "observer", None)
+        if observer:
+            pass  # Already tracking.  Keep tracking.
```
bcoe, line 257: that answers my question about calling it twice.

synthtool/metadata.py
```diff
@@ -224,38 +224,110 @@ def get_touched_file_paths(self) -> List[str]:
...
+            self.observer.schedule(self.handler, str(watch_dir), recursive=True)
+            self.observer.start()
+
+    def stop_tracking(self):
+        """Shut down existing watchdog observer.  Does nothing if not observing."""
+        observer = getattr(self, "observer", None)
+        if observer:
+            time.sleep(2)  # Finish collecting observations about modified files.
+            observer.stop()
+            observer.join()
```
bcoe, line 272: If I'm following this correctly, the handler runs in its own thread, and `join` will basically bring it in to our current thread allowing us access to the list it has populated?

bcoe (approved):
This is looking good to me 👍 

One nice to have, it could be good to have a method we can call other than `node_library` to start tracking, as there's a handful of libraries that use `synthtool` for templates, but don't use `protos` (_they call the template helper, but not `node_library`_).

PR #830, opened 2020-10-26
feat(java): Add support for googleapis-discovery repository

vam-google:
Synthtool with the changes from this PR was used to generate the initial version of java-compute DIREGAPIC client: https://github.com/googleapis/java-compute/tree/diregapic (not to be confused with discogapic - they are fundamentally different clients).


synthtool/gcp/gapic_bazel.py
```diff
@@ -274,6 +285,22 @@ def _clone_googleapis_private(self):
 
         return self._googleapis_private
 
+    def _clone_googleapis_discovery(self):
```
SurferJeffAtGoogle, line 288: Please add a test that exercises this new code path.

SurferJeffAtGoogle (changes requested):

synthtool/languages/java.py
```diff
@@ -199,7 +199,8 @@ def _common_generation(
     s.copy(
         [library / f"grpc-google-{cloud_prefix}{service}-{version}{suffix}/src"],
         f"grpc-google-{cloud_prefix}{destination_name}-{version}/src",
-        required=True,
+        # For REST-only clients, like java-compute, gRPC artifact does not exist
+        # required=True,
```
chingor13, line 203: Can we preserve the default behavior and override with optional arguments? This is a nice confidence check to ensure that we didn't typo something (identifying a no-op vs. not copying files).

chingor13 (changes requested):

SurferJeffAtGoogle:
@vam-google There has been no activity in this pull request for over 30 days.  What's your plan?

synthtool/languages/java.py
vam-google: Made it a function of diregapic parameter.

synthtool/gcp/gapic_bazel.py
vam-google: Looks like this whole file has never had tests. All of the previous 11 PRs to this file never had tests. Adding test for this little piece will require essentially creating the tests for everything else in the file. 

It would be a right thing to do, but given that most probably very soon we will not need this file at all (to be replaced with `googleapis-gen` stuff) and that my Python skills are not on par with what would be requried to write tests for such a sophisticated file, I'm asking for permission to skip writing test for it.

vam-google:
@chingor13 @SurferJeffAtGoogle PTAL

vam-google:
@SurferJeffAtGoogle I plan to merge it ASAP. I did not pay much attention to this PR because it was not blocking anyone. Now that we are about to regenerate new versions of Java, Python and PHP clients from googleapis-discovery this PR becomes essential.

synthtool/languages/java.py
```diff
@@ -260,6 +262,7 @@ def gapic_library(
         library=library,
         package_pattern=package_pattern,
         destination_name=destination_name,
+        diregapic=kwargs.get("diregapic", False),
```
chingor13, line 265: We can add to the named param list of `gapic_library()` and `bazel_library()` to avoid the `kwargs.get()`

synthtool/languages/java.py
```diff
@@ -306,6 +309,7 @@ def bazel_library(
         suffix="-java",
         destination_name=destination_name,
         cloud_api=cloud_api,
+        diregapic=kwargs.get("diregapic", False),
```
chingor13, line 312: We can add to the named param list of `gapic_library()` and `bazel_library()` to avoid the `kwargs.get()`

chingor13 (changes requested):
nit on the kwargs usage

chingor13 (approved):

synthtool/gcp/gapic_bazel.py
SurferJeffAtGoogle: Ok.

SurferJeffAtGoogle (approved):

PR #836, opened 2020-11-03
docs(node): update contributing instructions

freelerobot:


lesv (approved):

lesv (commented):
You don't always need an API key

synthtool/gcp/templates/node_library/CONTRIBUTING.md
```diff
@@ -39,6 +39,16 @@ accept your pull requests.
 
 ## Running the tests
 
+1.  To run the tests locally, you will need a Google Cloud Project with a [service account](https://cloud.google.com/docs/authentication/production#cloud-console) and an [API key](https://cloud.google.com/docs/authentication/api-keys)
+
+1.  Set the required environment variables. 
+```
+export GCLOUD_TESTS_PROJECT_ID=YOUR_GCP_PROJECT_ID
```
bcoe, line 46: it seems to reset the list in the markdown here when I preview, it might do the right thing if we indent:

```
1.  Set the required environment variables. 
```
export GCLOUD_TESTS_PROJECT_ID=YOUR_GCP_PROJECT_ID
 

 
it seems to reset the list in the markdown here when I preview, it might do the right thing if we indent:

```

```md
1.  Set the required environment variables. 
    \`\`\`
    export GCLOUD_TESTS_KEY=PATH_TO_YOUR_SERVICE_ACCOUNT_KEY.json
    export GCLOUD_TESTS_PROJECT_NUMBER=YOUR_GCP_PROJECT_NUMBER
    export GCLOUD_TESTS_API_KEY=YOUR_API_KEY
   \`\`\`
```

nicoleczhu: oops. Great catch, thank you.

synthtool/gcp/templates/node_library/CONTRIBUTING.md
```diff
@@ -39,6 +39,16 @@ accept your pull requests.
...
+export GCLOUD_TESTS_KEY=PATH_TO_YOUR_SERVICE_ACCOUNT_KEY.json
+export GCLOUD_TESTS_PROJECT_NUMBER=YOUR_GCP_PROJECT_NUMBER
```
bcoe, line 48: for most unit tests I think we only require that folks set a:

`GCLOUD_PROJECT` and a `GOOGLE_APPLICATION_CREDENTIALS`  (I think there are a few projects that require project number, but they're outliers).

bcoe (changes requested):
I think it's relatively rare that we ned to set the PROJECT_NUMBER for tests (_I think some APIs do require it, but only a couple_).

Perhaps we could get away with just pointing folks towards [these docs](https://cloud.google.com/docs/authentication/getting-started), but then adding a blurb about the fact that most tests require `GCLOUD_PROJECT` to be set? we could then point out that tests sometimes take additional environment variables, which should be documented at the top of each test file.

synthtool/gcp/templates/node_library/CONTRIBUTING.md
nicoleczhu: `we could then point out that tests sometimes take additional environment variables, which should be documented at the top of each test file.`
That's a great convention to follow. I don't think it's currently the majority case (at least for logging/errorreporting). Do you have a sample repo that already does this, so I can follow the formatting?

bcoe: I noticed this approach being used in `automl`:

https://github.com/googleapis/nodejs-automl/blob/master/samples/test/batch_predict.test.js#L23

Which has quite a few environment variables different than other libraries. Most libraries once a library is turned on on the account, just need the `GCLOUD_PROJECT` + `GOOGLE_APPLICATION_CREDENTIALS`.

sofisl:
> Perhaps we could get away with just pointing folks towards [these docs](https://cloud.google.com/docs/authentication/getting-started), but then adding a blurb about the fact that most tests require `GCLOUD_PROJECT` to be set? we could then point out that tests sometimes take additional environment variables, which should be documented at the top of each test file.

I agree with the statement above....these tests are weird, some require an env variable, some don't (like nodejs-dlp). I think the instructions make sense, as long as we put some kind of disclaimer to the effect of "these instructions may vary from library to library"...

SurferJeffAtGoogle:
Closing because this PR has been dormant for 90+ days.  Feel free to re-open if you intend to revise and merge.

PR #841, opened 2020-11-05
fix: export clients as both const and type

alexander-fenster:
The generator now generates the same code (both `const` and `type` to allow declaration merging), so we need to update it here as well. The idea is to have the same `src/index.ts` as in https://github.com/googleapis/nodejs-secret-manager/pull/184/files.

### DIFF
--- a/tests/fixtures/node_templates/index_samples/sample_index.ts
+++ b/tests/fixtures/node_templates/index_samples/sample_index.ts
@@ -20,6 +20,7 @@ import * as v1 from './v1';
 import * as v1beta1 from './v1beta1';
 
 const TextToSpeechClient = v1.TextToSpeechClient;
+type TextToSpeechClient = v1.TextToSpeechClient;
 
 export {v1, v1beta1, TextToSpeechClient};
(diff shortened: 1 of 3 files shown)


tests/fixtures/node_templates/index_samples/sample_index.ts
```diff
@@ -20,6 +20,7 @@ import * as v1 from './v1';
 import * as v1beta1 from './v1beta1';
 
 const TextToSpeechClient = v1.TextToSpeechClient;
+type TextToSpeechClient = v1.TextToSpeechClient;
```
bcoe, line 23: I didn't know this was a thing.

bcoe (approved):

tests/fixtures/node_templates/index_samples/sample_index.ts
alexander-fenster: Called "declaration merging". Tested [here](https://github.com/googleapis/nodejs-speech/pull/643), fixes [this](https://github.com/googleapis/gapic-generator-typescript/issues/694).

PR #856, opened 2020-11-18
docs(python): update intersphinx for grpc and auth

busunkim96:


synthtool/gcp/templates/python_library/docs/conf.py.j2
```diff
@@ -346,12 +346,12 @@ texinfo_documents = [
 # Example configuration for intersphinx: refer to the Python standard library.
 intersphinx_mapping = {
     "python": ("http://python.readthedocs.org/en/latest/", None),
```
tswast, line 348: Python is also redirecting to the https endpoint.

tswast (approved):

synthtool/gcp/templates/python_library/docs/conf.py.j2
busunkim96: Thanks Tim!

PR #864, opened 2020-12-01
chore: Update publish_javadoc.sh

lesv:
* add `export` to the envVar's to make sure maven knows about them.
* report the Java version

@tbpg FYI

### DIFF
--- a/synthtool/gcp/templates/java_library/.kokoro/release/publish_javadoc.sh
+++ b/synthtool/gcp/templates/java_library/.kokoro/release/publish_javadoc.sh
@@ -38,8 +38,8 @@ python3 -m pip install gcp-docuploader
 # compile all packages
 mvn clean install -B -q -DskipTests=true
 
-NAME={{ metadata['repo']['distribution_name'].split(':')|last }}
-VERSION=$(grep ${NAME}: versions.txt | cut -d: -f3)
+export NAME={{ metadata['repo']['distribution_name'].split(':')|last }}
+export VERSION=$(grep ${NAME}: versions.txt | cut -d: -f3)
 
 # build the docs
 mvn site -B -q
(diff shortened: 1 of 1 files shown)


synthtool/gcp/templates/java_library/.kokoro/release/publish_javadoc.sh
```diff
@@ -59,6 +59,9 @@ python3 -m docuploader upload . \
 
 popd
 
+# V2 currently requires Java 8 ONLY.
```
elharo, line 62: Could you check for Java 8 instead of printing the version?

lesv: Why?  It's just there to remind someone when they run across a problem.

lesv: PTAL

synthtool/gcp/templates/java_library/.kokoro/release/publish_javadoc.sh
```diff
@@ -59,7 +59,8 @@ python3 -m docuploader upload . \
 
 popd
 
-# V2
+# V2 due to problems w/ the released javadoc plugin doclava, Java 8 is required.  Beware of accidental updates.
```
elharo, line 62: should we be locking the version of the plugin in the pom.xml then? 

Is "doclava" correct?

lesv: It's done.  DocLava hasn't been published (by us) in 6 years.  I'm trying to get this released so we can the CloudRAD done.  We are also working on a replacement.

elharo: OK. I just never encountered that before. I'm still having trouble parsing this comment. It might be worth rewriting.

elharo (approved):

SurferJeffAtGoogle (approved):

Issue #431, opened 2020-03-09
Fix samples readme reference


jsoref:
Thanks for stopping by to let us know something could be better!

#### Environment details

  - `synthtool` version: master

#### Steps to reproduce

  1. Read the sentence in https://github.com/googleapis/nodejs-dlp/blob/master/README.md#samples
  2. Determine that the grammar is confusing and that the message is wrong.
  3. File https://github.com/googleapis/nodejs-dlp/pull/423
  4. Told [to file an issue](https://github.com/googleapis/nodejs-dlp/pull/423#issuecomment-596765256)

jsoref:
@bcoe:
Afaiu, you're thinking about https://github.com/googleapis/synthtool/blob/master/synthtool/gcp/templates/node_library/samples/README.md which is a separate file.

If you mean that the message itself is talking about those files, then the change should be of the form:
```patch
## Samples

-Samples are in the [`samples/`](https://github.com/{{  metadata['repo']['repo'] }}/tree/master/samples) directory. The samples' `README.md`
-has instructions for running the samples.
+Samples are in the [`samples/`](https://github.com/{{  metadata['repo']['repo'] }}/tree/master/samples) directory. Each sample's `README.md`
+has instructions for running its sample.
```

I'm happy to make a PR for that if that's the intention. Or if the intention is closer to my initial interpretation, I'm happy to make a PR for that.

sofisl:
Addressed in #872, will hold off on merging until after fix-it week

PR #872, opened 2020-12-07
docs: add instructions for authenticating for system tests

sofisl:
Addresses this docs question: https://github.com/googleapis/nodejs-googleapis-common/issues/327
We need to add authentication instructions when running system tests as well - these docs add to that.


JustinBeckwith (approved):

bcoe (approved):
This is looking good to me, mind running `synthtool` against one repo manually first, so we can make sure it looks right?

sofisl:
Can confirm it looks right! Also added in a fix to this: https://github.com/googleapis/synthtool/issues/431

bcoe (approved):

PR #873, opened 2020-12-08
fix(autosynth): apiary provider should include admin clients

chingor13:
Fixes https://github.com/googleapis/google-api-java-client-services/issues/6646

### DIFF
--- a/tests/test_autosynth_apiary_provider.py
+++ b/tests/test_autosynth_apiary_provider.py
@@ -43,3 +43,14 @@ def test_list_all_libraries_admin_snowflakes(mock_list_files):
     apis = apiary.list_apis()
     assert len(apis) == 1
     assert apis["admin"] == ["directory_v1", "reports_v1"]
+
+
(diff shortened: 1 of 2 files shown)


tests/test_autosynth_apiary_provider.py
```diff
@@ -0,0 +1,45 @@
...
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import os
+from unittest.mock import patch
+
+from autosynth.github import GitHub
+from autosynth.providers import apiary
+
+
```
SurferJeffAtGoogle, line 21: Could you also add negative tests: confirm some strings do not match the regex?

chingor13: Done. We relaxed the pattern here, so any .json with at least 2 segments matches. This seems fine for now as there's currently nothing else in the discoveries/ folder of discovery-artifact-manager.

SurferJeffAtGoogle (approved):

PR #876, opened 2020-12-12
feat: a docker image that works for Java and Nodejs

SurferJeffAtGoogle:
The kokoro configs haven't been changed to invoke the docker image.
That will arrive in a later pull request.

### DIFF
--- a/.kokoro-autosynth/build-with-docker.sh
+++ b/.kokoro-autosynth/build-with-docker.sh
@@ -25,6 +25,8 @@ docker run \
     --env MULTISYNTH_CONFIG=${MULTISYNTH_CONFIG} \
     --env MULTISYNTH_SHARD=${MULTISYNTH_SHARD} \
     --env SYNTHTOOL_TRACK_OBSOLETE_FILES=${SYNTHTOOL_TRACK_OBSOLETE_FILES} \
+    `# Kokoro exposes this as a file, but the scripts expect just a plain variable.` \
+    --env GITHUB_TOKEN=$(cat ${KOKORO_KEYSTORE_DIR}/73713_yoshi-automation-github-key)
     -v ${KOKORO_ARTIFACTS_DIR}:/kokoro/artifacts \
     -v ${KOKORO_KEYSTORE_DIR}:/kokoro/keystore \
     gcr.io/cloud-devrel-kokoro-resources/synthtool:latest /synthtool/docker/entrypoint.sh
--- a/.kokoro-autosynth/build-with-docker.sh
+++ b/.kokoro-autosynth/build-with-docker.sh
@@ -1,0 +1,30 @@
+#!/bin/bash
+# Copyright 2020 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     http://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -eo pipefail
+
+docker run \
+    --env AUTOSYNTH_MULTIPLE_COMMITS=${AUTOSYNTH_MULTIPLE_COMMITS} \
+    --env AUTOSYNTH_MULTIPLE_PRS=${AUTOSYNTH_MULTIPLE_PRS} \
+    --env GITHUB_EMAIL=${GITHUB_EMAIL} \
+    --env GITHUB_USER=${GITHUB_USER} \
+    --env KOKORO_ARTIFACTS_DIR=/kokoro/artifacts \
+    --env KOKORO_KEYSTORE_DIR=/kokoro/keystore \
+    --env MULTISYNTH_CONFIG=${MULTISYNTH_CONFIG} \
+    --env MULTISYNTH_SHARD=${MULTISYNTH_SHARD} \
+    --env SYNTHTOOL_TRACK_OBSOLETE_FILES=${SYNTHTOOL_TRACK_OBSOLETE_FILES} \
+    -v ${KOKORO_ARTIFACTS_DIR}:/kokoro/artifacts \
+    -v ${KOKORO_KEYSTORE_DIR}:/kokoro/keystore \
+    gcr.io/cloud-devrel-kokoro-resources/synthtool:latest /synthtool/docker/entrypoint.sh
--- a/cloudbuild.yaml
+++ b/cloudbuild.yaml
@@ -4,7 +4,6 @@ steps:
         '-t', 'gcr.io/cloud-devrel-kokoro-resources/synthtool:$SHORT_SHA',
         '-t', 'gcr.io/cloud-devrel-kokoro-resources/synthtool:latest', 
         '-f', 'docker/Dockerfile', '.' ]
-    - name: 'gcr.io/cloud-builders/docker'
-      args: ['push', 'gcr.io/cloud-devrel-kokoro-resources/synthtool:$SHORT_SHA']
-    - name: 'gcr.io/cloud-builders/docker'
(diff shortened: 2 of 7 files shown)


docker/Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+# The "npm config set unsafe-perm true" is necessary because npm will refuse to perform
+# some operations if the user is running as root.  The root user of a docker image
+# doesn't have root permissions on the machine, so it's safe to execute them.
+# I learned this from https://geedew.com/What-does-unsafe-perm-in-npm-actually-do/
+RUN /bin/bash -c "source /root/.nvm/nvm.sh && nvm install --lts && npm config set unsafe-perm true"
+
+####################### Install bazel 3.0.0
+RUN mkdir -p /bazel
+RUN curl -L https://github.com/bazelbuild/bazel/releases/download/3.0.0/bazel-3.0.0-linux-x86_64 -o /bazel/bazel
+RUN chmod +x /bazel/bazel
```
chingor13, line 23: If we're worried about optimizing the image size (number of layers), we should combine `RUN` statements with `&&`.

docker/Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+# Put synthtool in the PYTHONPATH so we can invoke it with `python -m autosynth`.
+ENV PYTHONPATH="/synthtool"
+
+# Copy synthtool.
+COPY synthtool /synthtool/synthtool
+COPY autosynth /synthtool/autosynth
+COPY docker /synthtool/docker
+
+# By default, display instructions on how to run synthtool.
+CMD ["/bin/cat", "/synthtool/docker/README.txt"]
```
chingor13, line 49: Generally, you'll want to set `ENTRYPOINT`, but we should consider how we want to execute this container. Perhaps consider creating a bash script to serve as an entrypoint which can print usage, or execute autosynth or synthtool depending on the command passed to the docker run command.

docker/entrypoint.sh
```diff
@@ -0,0 +1,37 @@
...
+
+# Disable buffering, so that the logs stream through.
+export PYTHONUNBUFFERED=1
+
+# Add github to known hosts.
+mkdir -p ~/.ssh
+ssh-keyscan github.com >> ~/.ssh/known_hosts
+
+# Kokoro exposes this as a file, but the scripts expect just a plain variable.
+export GITHUB_TOKEN=$(cat ${KOKORO_KEYSTORE_DIR}/73713_yoshi-automation-github-key)
```
chingor13, line 31: Consider abstracting this and exposing a nicer env var to the docker run command.

docker/entrypoint.sh
```diff
@@ -0,0 +1,37 @@
...
+source /root/.bashrc
+
+cd /synthtool
+
+# Disable buffering, so that the logs stream through.
+export PYTHONUNBUFFERED=1
+
+# Add github to known hosts.
+mkdir -p ~/.ssh
+ssh-keyscan github.com >> ~/.ssh/known_hosts
```
chingor13, line 28: Can this be done at image build time?

chingor13 (changes requested):

docker/Dockerfile
SurferJeffAtGoogle: I'm not yet worried.  What symptoms will I observe when there are too many layers?

docker/Dockerfile
SurferJeffAtGoogle: CMD vs ENTRYPOINT is a subtlety I don't fully grasp.

I read [this article](https://www.ctl.io/developers/blog/post/dockerfile-entrypoint-vs-cmd/) and concluded CMD was better, because I want to keep the command line as simple as possible when the user actually invokes the docker image, and the command line is simpler without an `--entrypoint`.

docker/entrypoint.sh
SurferJeffAtGoogle: It could, but then we'd have to rebuild docker images when github.com changes its keys.

cloudbuild.yaml
```diff
@@ -0,0 +1,10 @@
+steps:
+    - name: 'gcr.io/cloud-builders/docker'
+      args: [ 'build',
+        '-t', 'gcr.io/cloud-devrel-kokoro-resources/synthtool:$SHORT_SHA',
+        '-t', 'gcr.io/cloud-devrel-kokoro-resources/synthtool:latest', 
+        '-f', 'docker/Dockerfile', '.' ]
+    - name: 'gcr.io/cloud-builders/docker'
+      args: ['push', 'gcr.io/cloud-devrel-kokoro-resources/synthtool:$SHORT_SHA']
+    - name: 'gcr.io/cloud-builders/docker'
+      args: ['push', 'gcr.io/cloud-devrel-kokoro-resources/synthtool:latest']
```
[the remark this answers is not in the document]
SurferJeffAtGoogle: Done.

docker/Dockerfile
```diff
@@ -0,0 +1,49 @@
...
+RUN curl -L https://github.com/bazelbuild/bazel/releases/download/3.0.0/bazel-3.0.0-linux-x86_64 -o /bazel/bazel
+RUN chmod +x /bazel/bazel
+ENV PATH="/bazel:${PATH}"
+
+# bazel requires zip.
+RUN apt-get update
+RUN apt-get install zip
+
+###################### Install java
+RUN curl -L https://download.java.net/java/GA/jdk15.0.1/51f4f36ad4ef43e39d0dfdbaf6549e32/9/GPL/openjdk-15.0.1_linux-x64_bin.tar.gz -o java.tar.gz
```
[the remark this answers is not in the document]
SurferJeffAtGoogle: I searched our build logs for an indication of which version of the jdk is currently being used, but found no clues.  Is the version easily available to you?

docker/entrypoint.sh
SurferJeffAtGoogle: Done.

docker/Dockerfile
chingor13: This really only matters if depending on how we plan to use the image. If it's a runtime to run a bunch of shell commands, then we can just set a default CMD. If we want to use it like binary (e.g. run `docker gcr.io/.../synthtool extra args here`), then entrypoint is probably preferred as extra args are appended to the entrypoint.

docker/Dockerfile
chingor13: Bigger image sizes and longer downloads (each layer is pulled and decompressed).

docker/Dockerfile
SurferJeffAtGoogle: I imagine people running shell commands.

.kokoro-autosynth/build-with-docker.sh
```diff
@@ -0,0 +1,32 @@
...
+
+set -eo pipefail
+
+docker run \
+    --env AUTOSYNTH_MULTIPLE_COMMITS=${AUTOSYNTH_MULTIPLE_COMMITS} \
+    --env AUTOSYNTH_MULTIPLE_PRS=${AUTOSYNTH_MULTIPLE_PRS} \
+    --env GITHUB_EMAIL=${GITHUB_EMAIL} \
+    --env GITHUB_USER=${GITHUB_USER} \
+    --env KOKORO_ARTIFACTS_DIR=/kokoro/artifacts \
+    --env KOKORO_KEYSTORE_DIR=/kokoro/keystore \
```
chingor13, line 24: Is this intentional to be hard-coding these for Kokoro?

docker/Dockerfile
```diff
@@ -0,0 +1,49 @@
+# build from the root of this repo:
```
chingor13, line 1: This Dockerfile probably needs a license header.

docker/Dockerfile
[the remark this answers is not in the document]
chingor13: Did we want to switch this one to Java 8?

chingor13 (changes requested):

bcoe:
@SurferJeffAtGoogle apologize for dropping this review on the ground. Is this work still pertinent, with the work we're doing around OwlBot; once concern is the potential need to run docker in docker?

SurferJeffAtGoogle:
I have no immediate plan to merge this PR, so I'll close it.

PR #883, opened 2020-12-16
feat(java): add helper to remove methods in post processing

chingor13:
Microgenerator cannot be configured to support legacy resource name classes. We need to remove a few methods to support backward-compatibility.


stephaniewang526 (approved):

stephaniewang526:
Thank you for the change! This should unblock microgenerator migration for bigquerydatatransfer: https://github.com/googleapis/java-bigquerydatatransfer/pull/570 (currently outstanding due to the fact that we need to remove 2 methods post-processing).

synthtool/languages/java.py
```diff
@@ -398,3 +399,43 @@ def custom_templates(files: List[str], **kwargs) -> None:
...
+
+
+def remove_method(filename: str, signature: str):
+    """Helper to remove an entire method.
+
+    Goes line-by-line to detect the start of the block. Determines
+    the end of the block by a closing brace at the same indentation
+    level. This requires the file to be correctly formatted.
+    """
+    lines = []
```
SurferJeffAtGoogle, line 411: I think I can write some Java code that would be misinterpreted by this remove_method() function.

SurferJeffAtGoogle (changes requested):
We're trying to remove code that semantically modifies the code coming out of the microgenerator.  Please file a bug or feature request on the microgenerator describing your need.

CC @vam-google

chingor13:
cc @miraleung

This is to handle ~4 GA clients with legacy resource name implementations. This override was specified in the gapic yaml which is no longer possible for any microgenerator. When these client next release a breaking change, we want to remove these legacy resource names and it will be difficult if these 4 cases are hard-coded in the microgenerator.

synthtool/languages/java.py
chingor13: I'm not convinced if the java file is formatted correctly (with google-java-format).

miraleung:
The monolith used to generate resource name classes that subclassed other generated ones, but that was removed before the microgenerator came around. Subsequently, the subclassed resource name classes were deprecated.

However, they contain some `public static` methods that are part of the superclass as well. Since this special case applies only to a very few number of APIs and is deprecated, removing them (in favor of the superclass's method) is the optimal temporary fix, as opposed to hardcoding a significant amount of complex logic in the microgenerator.

miraleung:
I agree that semantic modification can be error-prone, which also motivates moving such logic alterations upstream for OneHorizontal. As semantic changes are already in-scope for synthtool's existing functionality, this doesn't seem to deviate from the current post-processing model. 

We could remove this feature after those 4 APIs have taken a major version bump (unless we need to support regeneration of the older versions forever).

SurferJeffAtGoogle:
Please copy, paste, and continue this discussion in a bug, and add the bug to hot list 2887527.

synthtool/languages/java.py
SurferJeffAtGoogle: Describe what the `signature` parameter is supposed to look like in the doc comment.

chingor13:
> Please copy, paste, and continue this discussion in a bug, and add the bug to hot list 2887527.

Done. b/176829483

synthtool/languages/java.py
chingor13: Done.

SurferJeffAtGoogle: What if there are so many arguments to the function that the signature wraps to two lines?

chingor13: It'll only need to match the first line. We could extend this to handle multiple lines, but as it is right now, there is no need.

SurferJeffAtGoogle: Ok.  In the doc comment, please show an example of what this function is designed to match and remove.

stephaniewang526:
@SurferJeffAtGoogle is there anything blocking the merging of this PR please?

SurferJeffAtGoogle (approved):

SurferJeffAtGoogle (approved):

PR #923, opened 2021-01-26
feat(templates): pass "template" positional to run templates only

bcoe:
Allow templates to be run in isolation, via:

```
synthtool template
```

This is in prep for separating this logic out from the gapic generation logic.


autosynth/change_pusher.py
```diff
@@ -100,10 +100,7 @@ def push_changes(
             )
         else:
             pr = self._gh.create_pull_request(
-                self._repository,
-                branch=branch,
-                title=pr_title[0:250],
-                body=new_body,
+                self._repository, branch=branch, title=pr_title[0:250], body=new_body,
```
bcoe, line 103: these formatting changes were automatic as a result of running `nox` locally.

synthtool/__main__.py
```diff
@@ -54,6 +55,22 @@ def extra_args() -> List[str]:
     return list(_extra_args)
 
 
+@click.command()
+@click.version_option(message="%(version)s", version=VERSION)
+@click.argument("command")
+@click.argument("language", default="node")
+@click.argument("extra_args", nargs=-1)
+def template(command: str, language: str, extra_args: Sequence[str]):
```
bcoe, line 63: actual code change is adding an alternate command here.

synthtool/__main__.py
```diff
@@ -19,9 +19,10 @@
 
 import click
 import pkg_resources
+import synthtool.gcp as gcp
 from synthtool.log import logger
 import synthtool.metadata
-from synthtool import preconfig
+from synthtool import preconfig, copy
```
chingor13, line 25: you could add the `gcp` import here too.

```suggestion
from synthtool import copy, gcp, preconfig
```

synthtool/__main__.py
```diff
@@ -54,6 +55,22 @@ def extra_args() -> List[str]:
...
+@click.argument("extra_args", nargs=-1)
+def template(command: str, language: str, extra_args: Sequence[str]):
+    f"""Run templates for common repository files, e.g., CONTRIBUTING.md
+
+    {preconfig.PRECONFIG_HELP}
+    """
+    common_templates = gcp.CommonTemplates()
+    if language == "node":
+        templates = common_templates.node_library()
+        copy([templates])
```
chingor13, line 71: Add an `else` clause with a warning that nothing else is supported?

synthtool/__main__.py
```diff
@@ -67,35 +84,40 @@ def main(synthfile: str, metadata: str, extra_args: Sequence[str]):
...
-      AUTOSYNTH_USE_SSH:         Access github repos via ssh instead of https.
-      {preconfig.PRECONFIG_ENVIRONMENT_VARIABLE}:  Path to a json file.
+    AUTOSYNTH_USE_SSH:         Access github repos via ssh instead of https.
+    {preconfig.PRECONFIG_ENVIRONMENT_VARIABLE}:  Path to a json file.
 
 
     {preconfig.PRECONFIG_HELP}
     """
-    _extra_args.extend(extra_args)
+    if synthfile == "template":
```
chingor13, line 97: I'm not sure how this works with click and executing as a module, but this feels wrong.

Perhaps we should add this feature as `python3 -m synthtool.templates` rather than overloading the main module execution.

chingor13 (changes requested):

synthtool/__main__.py
bcoe: So, add a `__main__.py` to templates?

I was trying to figure out how to make click handle a default command plus other commands, and it wasn't immediately obvious.

SurferJeffAtGoogle: I agree that we should put this command in its own .py file, and invoke it directly from there.

I like to see verbs in command names.  What is synthtool doing with the templates?  Create a file <verb_templates>.py with its own main function and move this code in there.

bcoe:
upon further investigation, I think we'll use a slightly different approach to generation out of the gate.

PR #930, opened 2021-01-28
build(java): generate docfx yml on release

eaball35:


synthtool/gcp/templates/java_library/.kokoro/release/publish_javadoc11.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+# Copyright 2019 Google Inc.
```
lesv, line 2: This is new, should be 2021 Google LLC

lesv (approved):
Please fix the copyright in your script, then LGTM

synthtool/gcp/templates/java_library/.kokoro/release/publish_javadoc11.cfg
```diff
@@ -0,0 +1,30 @@
...
+
+# Configure the docker image for kokoro-trampoline
+env_vars: {
+  key: "TRAMPOLINE_IMAGE"
+  value: "gcr.io/cloud-devrel-kokoro-resources/java11"
+}
+
+env_vars: {
+  key: "TRAMPOLINE_BUILD_FILE"
+  value: "github/java-memcache/.kokoro/release/publish_javadoc11.sh"
```
chingor13, line 17: If the 2 scripts are the same and the only differences are the docker image and the bucket name, we can re-use the same shell script.

chingor13: Actually it looks like the `destination-prefix` option is new. You could potentially make that configurable via an env variable or keep the 2 separate scripts.

eaball35: Makes sense, I think I'll just leave as is unless you have a strong opinion about this

PR #944, opened 2021-02-09
feat: add googleapis-gen as a tracked source

chingor13:
Adds the ability to copy pregenerated code directly from `googleapis/googleapis-gen` rather than building it at runtime via bazel. This correctly can track the git source in synth.metadata and should allow autosynth to correctly binary search through history for context-aware-commits (googleapis and googleapis-gen have a 1:[0-1] relationship so history should be preserved).

Example `synth.py` for Java:

```python3
import synthtool.languages.java as java

service = 'domains'
versions = ['v1beta1']

for version in versions:
  java.pregenerated_library(
    service=service,
    version=version,
    path=f'google/cloud/{service}/{version}',
  )

java.common_templates()
```

### DIFF
--- a/synthtool/gcp/pregenerated.py
+++ b/synthtool/gcp/pregenerated.py
@@ -13,7 +13,7 @@
 # limitations under the License.
 
 from pathlib import Path
-from typing import Optional, Union
+from typing import Optional
 import os
 
 from synthtool.log import logger
--- a/synthtool/gcp/pregenerated.py
+++ b/synthtool/gcp/pregenerated.py
@@ -19,20 +19,20 @@ import os
 from synthtool.log import logger
 from synthtool.sources import git
 
-GOOGLEAPIS_GEN_URL: str = git.make_repo_clone_url("googleapis/googleapis-gen")
-LOCAL_GOOGLEAPIS_GEN: Optional[str] = os.environ.get("SYNTHTOOL_GOOGLEAPIS_GEN")
-
 
 class Pregenerated:
     """A synthtool component that copies pregenerated bazel code."""
 
     def __init__(self):
-        if LOCAL_GOOGLEAPIS_GEN:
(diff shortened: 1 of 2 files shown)


synthtool/gcp/pregenerated.py
```diff
@@ -0,0 +1,38 @@
...
+    def __init__(self):
+        if LOCAL_GOOGLEAPIS_GEN:
+            self._googleapis_gen = Path(LOCAL_GOOGLEAPIS_GEN).expanduser()
+            logger.debug(f"Using local googleapis-gen at {self._googleapis_gen}")
+        else:
+            logger.debug("Cloning googleapis-gen.")
+            self._googleapis_gen = git.clone(GOOGLEAPIS_GEN_URL)
+
+    def generate(self, path: str) -> Path:
+        return self._googleapis_gen / path
```
chingor13, line 38: Something to consider is creating a temp directory and copying the files into the that temp directory. It's possible that the code that calls this will modify the files in the directory returned.

busunkim96 (approved):

synthtool/gcp/pregenerated.py
```diff
@@ -0,0 +1,38 @@
...
+
+from pathlib import Path
+from typing import Optional
+import os
+
+from synthtool.log import logger
+from synthtool.sources import git
+
+GOOGLEAPIS_GEN_URL: str = git.make_repo_clone_url("googleapis/googleapis-gen")
+LOCAL_GOOGLEAPIS_GEN: Optional[str] = os.environ.get("SYNTHTOOL_GOOGLEAPIS_GEN")
```
SurferJeffAtGoogle, line 23: I see no advantage to making these globals.

chingor13: Done

SurferJeffAtGoogle (approved):

PR #948, opened 2021-02-17
feat: a docker image for owlbot's nodejs post processor

SurferJeffAtGoogle:
Here's a sample owlbot.py that works with this docker image:

https://github.com/googleapis/nodejs-vision/blob/owlbot.py/owlbot.py
Observe that the call to `s.replace()` was not idempotent, so I had to add a conditional check to see if had already been applied.

### DIFF
--- a/docker/owlbot/nodejs/Dockerfile
+++ b/docker/owlbot/nodejs/Dockerfile
@@ -1,0 +1,51 @@
+# Copyright 2021 Google LLC
+#
+# Licensed under the Apache License, Version 2.0 (the "License");
+# you may not use this file except in compliance with the License.
+# You may obtain a copy of the License at
+#
+#     https://www.apache.org/licenses/LICENSE-2.0
+#
+# Unless required by applicable law or agreed to in writing, software
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+# build from the root of this repo:
+# docker build -t gcr.io/cloud-devrel-kokoro-resources/owlbot-nodejs -f docker/owlbot/nodejs/Dockerfile .
+FROM python:3.6.12-buster
+
+WORKDIR /
+
+######################## Install nodejs.
+# Use `nvm` to bootstrap the installation of node.js and npm.
+# To learn more: https://github.com/nvm-sh/nvm
+RUN curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.35.3/install.sh | bash
+
+# Use the latest LTS version of nodejs.  This can change over time, but
+# should prevent the need to modify this file every year.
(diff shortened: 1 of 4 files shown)


docker/owlbot/nodejs/Dockerfile
```diff
@@ -0,0 +1,51 @@
...
+# the live repo.
+ENV SYNTHTOOL_TEMPLATES="/synthtool/synthtool/gcp/templates"
+
+# Copy synthtool.
+COPY synthtool /synthtool/synthtool
+COPY autosynth /synthtool/autosynth
+COPY docker /synthtool/docker
+
+ENTRYPOINT [ "/bin/bash" ]
+CMD [ "/synthtool/docker/owlbot/nodejs/entrypoint.sh" ]
```
bcoe, line 51: Will this path be affected by us setting an alternate `dir` in cloud build [see](https://github.com/googleapis/repo-automation-bots/blob/master/packages/owl-bot/cloud-build/update-pr.yaml#L20).

Or are we protected by virtue of the fact that it's an absolute path?

docker/owlbot/nodejs/entrypoint.sh
```diff
@@ -0,0 +1,21 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+set -e
+# Doesn't get sourced in root shell, for reasons I don't understand.
+source "$HOME/.bashrc"
+set -x
+cd "$OWLBOT_WORKING_DIR"
```
bcoe, line 20: Currently we run the container like this:

```
  - name: ${_CONTAINER}
    dir: '${_REPOSITORY}'
```

It sounds like we'd want to se an additional environment variable equal to `OWLBOT_WORKING_DIR`?

bcoe (commented):
a couple questions.

docker/owlbot/nodejs/Dockerfile
SurferJeffAtGoogle: I reread the documentation about cloud build's `dir` parameter, and I don't think I grasp it 100%.

But according to my present understanding, yes, because they are absolute paths, they should work regardless of the current working directly.

docker/owlbot/nodejs/entrypoint.sh
SurferJeffAtGoogle: Actually, we can run this container the same way:

```yaml
  - name: ${_CONTAINER}
    dir: '${_REPOSITORY}'
```

because `$OWLBOT_WORKING_DIR` will be empty, and calling `cd` with no arguments yields no error and no change in directory.

But that triggers the question do we need OWLBOT_WORKING_DIR at all?  I suppose not.  I'll remove it.

SurferJeffAtGoogle:
Ready for another look.

bcoe (approved):

PR #952, opened 2021-02-18
feat(owlbot): build linter into node container

bcoe:
1. install linter when docker container is created.
2. introduce `postprocess_gapic_library_hermetic` which allows post processing to be run without an `npm install` step.

### DIFF
--- a/docker/owlbot/nodejs/Dockerfile
+++ b/docker/owlbot/nodejs/Dockerfile
@@ -51,7 +51,7 @@ COPY docker /synthtool/docker
 # Install dependencies used for post processing:
 # * gts/typescript are used for linting.
 # * google-gax is used for compiling protos.
-RUN /bin/bash -c "source /root/.nvm/nvm.sh && mkdir node_modules && npm i gts@3.1.0 google-gax@2.10.3 typescript@3.9.9"
+RUN /bin/bash -c "source /root/.nvm/nvm.sh && cd /synthtool && mkdir node_modules && npm i gts@3.1.0 google-gax@2.10.3 typescript@3.9.9"
 
 ENTRYPOINT [ "/bin/bash" ]
 CMD [ "/synthtool/docker/owlbot/nodejs/entrypoint.sh" ]
--- a/docker/owlbot/nodejs/Dockerfile
+++ b/docker/owlbot/nodejs/Dockerfile
@@ -46,7 +46,10 @@ ENV SYNTHTOOL_TEMPLATES="/synthtool/synthtool/gcp/templates"
 COPY synthtool /synthtool/synthtool
 COPY autosynth /synthtool/autosynth
 COPY docker /synthtool/docker
-RUN cd /synthtool; mkdir node_modules; npm i gts@3.1.0 google-gax@2.10.3
+# Install dependencies used for post processing:
+# * gts/typescript are used for linting.
+# * google-gax is used for compiling protos.
+RUN cd /synthtool; mkdir node_modules; npm i gts@3.1.0 google-gax@2.10.3 typescript@3.9.9
 
 ENTRYPOINT [ "/bin/bash" ]
 CMD [ "/synthtool/docker/owlbot/nodejs/entrypoint.sh" ]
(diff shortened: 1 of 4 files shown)


synthtool/languages/node.py
```diff
@@ -179,6 +180,22 @@ def fix(hide_output=False):
...
+    is already installed in a well known location on disk:
+    """
+    logger.debug("Copy eslint config")
+    shell.run(
+        ["cp", "-r", f"{_TOOLS_DIRECTORY}/node_modules", "."], hide_output=hide_output
+    )
+    logger.debug("Running fix...")
+    shell.run(
+        [f"{_TOOLS_DIRECTORY}/node_modules/.bin/gts", "fix", "src"],
+        hide_output=hide_output,
```
SurferJeffAtGoogle, line 195: If this fails, do we want the generation step to fail?  If so, add the argument `check=True`

bcoe: switched to check `True`.

docker/owlbot/nodejs/entrypoint.sh
```diff
@@ -14,8 +14,7 @@
 # limitations under the License.
 
 set -e
-# Doesn't get sourced in a login shell, according to bash's man page and confirmed
-# via experiment, so we have to source it manually.
-source "$HOME/.bashrc"
+source /root/.nvm/nvm.sh
```
bcoe, line 17: @SurferJeffAtGoogle it took some trial and error to figure out an appropriate `entrypoint.sh` when `dir` was set in Cloud Build, this approach is working, see:

https://github.com/bcoe/owlbot-testing/pull/14

docker/owlbot/nodejs/Dockerfile
```diff
@@ -46,6 +48,10 @@ ENV SYNTHTOOL_TEMPLATES="/synthtool/synthtool/gcp/templates"
 COPY synthtool /synthtool/synthtool
 COPY autosynth /synthtool/autosynth
 COPY docker /synthtool/docker
+# Install dependencies used for post processing:
+# * gts/typescript are used for linting.
+# * google-gax is used for compiling protos.
+RUN /bin/bash -c "source /root/.nvm/nvm.sh && cd /synthtool && mkdir node_modules && npm i gts@3.1.0 google-gax@2.10.3 typescript@3.9.9"
```
bcoe, line 54: @alexander-fenster when we want to update a dependency, we simply update this Dockerfile and it will kick off a build of the container.

synthtool/languages/node.py
```diff
@@ -188,9 +208,29 @@ def compile_protos(hide_output=False):
...
 
 
+def compile_protos_hermetic(hide_output=False):
+    """
+    Compiles protos into .json, .js, and .d.ts files using
+    compileProtos script from google-gax.
+    """
+    logger.debug("Compiling protos...")
+    shell.run(
+        [f"{_TOOLS_DIRECTORY}/node_modules/.bin/compileProtos", "src"],
```
bcoe, line 218: @alexander-fenster as a TODO, `compileProtos` is not actually hermetic, because it performs an install step.

It would be good if we could get rid of this install step, so that we're running exactly the version of `compileProtos` installed when the container is built.

docker/owlbot/nodejs/entrypoint.sh
SurferJeffAtGoogle: Was it failing without this change?

bcoe: It was failing with `source "$HOME/.bashrc"`, I believe because `$HOME` becomes `dir`.

It was also failing unless I sourced `source /root/.nvm/nvm.sh`, because of not being able to find the `npm` bin.

docker/owlbot/nodejs/entrypoint.sh
```diff
@@ -14,8 +14,7 @@
...
+export PYTHONPATH="/synthtool"
```
bcoe, line 18:
```suggestion
```

I'm betting that I potentially don't need `PYTHONPATH`.

SurferJeffAtGoogle (approved):

docker/owlbot/nodejs/entrypoint.sh
SurferJeffAtGoogle: That's a bet you'll lose!  It needs to be set as is.  Otherwise, your `owlbot.py` script will fail to import stuff from synthtool.

PR #958, opened 2021-02-19
fix: pregenerated code needs to return a copy

### LINKED ISSUE #957: Pre-generated code runs post-processor multiple times
Example: https://github.com/googleapis/java-analytics-data/pull/143

What's happening is autosynth runs synthtool multiple times. Other generators return new copies of generated code which can be freely modified because subsequent generations are generated into a new directory. Autosynth does not "reset" these generated directories for the next instance. Therefore, if the post-processing step is non-idempotent, then the fix may be applied multiple times.

chingor13:
Autosynth may not reset the git clone of a given source - other
generators return a temp directory which contains the generated
code.

This fixes the case where autosynth is running the post processing
extra times on the same generated files and the post processing
step is not idempotent (e.g. adding a license header).

Fixes #957

### DIFF
--- a/synthtool/gcp/pregenerated.py
+++ b/synthtool/gcp/pregenerated.py
@@ -14,6 +14,8 @@
 
 from pathlib import Path
 import os
+import shutil
+import tempfile
 
 from synthtool.log import logger
 from synthtool.sources import git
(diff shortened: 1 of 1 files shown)


chingor13:
Running autosynth on the affected result: https://github.com/googleapis/java-analytics-data/pull/149

bcoe (approved):

PR #960, opened 2021-02-24
chore: make updates to toc.yml before upload

eaball35:
- adding README as overview
- sorting toc on package
- adding product hierarchy
- script cleanup


eaball35:
Tested these changes out in single library (on branch in java-memcache) and is working correctly as expected

eaball35:
b/181142127
b/179920619
b/179799230

synthtool/gcp/templates/java_library/.kokoro/release/prepare_docs.py
```diff
@@ -0,0 +1,47 @@
+# Copyright 2021 Google Inc.
```
chingor13, line 1:
```suggestion
# Copyright 2021 Google LLC
```

chingor13 (changes requested):
Is this testable anywhere? Ideally, we can test this script (perhaps make this part of synthtool)

chingor13 (changes requested):
The CI failure is because the python template is being compiled in-place, then the template engine is confused because it sees a new compiled.py file and tries to load it as a utf-8 file (which it's not).

From an offline chat, we decided the python script belongs in a single place in a python library (gcp-docuploader/publisher perhaps?) rather than as a standalone python file in 100+ repos.

eaball35:
Moving this out of synthtool and into docuploader instead https://github.com/googleapis/docuploader/pull/58

PR #971, opened 2021-03-04
feat(node): add helper for detecting versions

bcoe:
Rather than hard coding the list of API versions available, detect based on folders in `./src` (this is accurate for most auto-generated Node.js libraries).


SurferJeffAtGoogle (approved):

docker/owlbot/nodejs/Dockerfile
```diff
@@ -12,7 +12,7 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
-# Version 0.1.1
+# Version 0.2.0
```
SurferJeffAtGoogle, line 15: Is anything consuming this version number?

SurferJeffAtGoogle (approved):

docker/owlbot/nodejs/Dockerfile
bcoe: No, it's just meant as a way to kick off the Cloud Build job. Eventually I'm picturing this metadata could be used upstream when the PR is created for the new docker container.

PR #987, opened 2021-03-16
feat(node): invoke owlbot_main() where there's no owlbot.py in the repo

SurferJeffAtGoogle:
This will allow us to remove synth.py _and_ owlbot.py from many nodejs repos.

### DIFF
--- a/synthtool/languages/node.py
+++ b/synthtool/languages/node.py
@@ -17,11 +17,13 @@ from jinja2 import FileSystemLoader, Environment
 from pathlib import Path
 import os
 import re
-from synthtool import shell
+from synthtool import _tracked_paths, gcp, shell, transforms
 from synthtool.gcp import samples, snippets
 from synthtool.log import logger
 from synthtool.sources import git
 from typing import Any, Dict, List
+import logging
+import shutil
 
 _REQUIRED_FIELDS = ["name", "repository"]
 _TOOLS_DIRECTORY = "/synthtool"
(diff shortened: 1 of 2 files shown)


synthtool/languages/node.py
```diff
@@ -260,3 +262,58 @@ def postprocess_gapic_library_hermetic(hide_output=False):
...
+    s_copy = transforms.move
+    if staging.is_dir():
+        # Load the default version defined in .repo-metadata.json.
+        default_version = json.load(open(".repo-metadata.json", "rt"))[
+            "default_version"
+        ]
+        # Collect the subdirectories of the staging directory.
+        versions = [v.name for v in staging.iterdir() if v.is_dir()]
+        # Reorder the versions so the default version always comes last.
+        versions = [v for v in versions if v != default_version] + [default_version]
```
bcoe, line 301: I like this lambda.

synthtool/languages/node.py
```diff
@@ -260,3 +262,58 @@ def postprocess_gapic_library_hermetic(hide_output=False):
...
+
+        # Copy each version directory into the root.
+        for version in versions:
+            library = staging / version
```
bcoe, line 305: Looks like `transforms.move` already did what we needed?

bcoe (approved):

synthtool/languages/node.py
SurferJeffAtGoogle: Yep.  It was called `s.copy()` in synth.py files.

PR #991, opened 2021-03-18
feat: add Java owl-bot post-processing image

chingor13:


SurferJeffAtGoogle (approved):
I see no calls to the synthtool library code at all.  Did I miss one?

If there are no calls to synthtool, then what do you think about creating a new repo just for java with this code and templates?  Ah, well, that may make migration difficult because then there would be two copies of the templates to keep in sync.

Anyway, migrating this code out of the synthtool repo would reduce your maintenance burden, so something to consider in the future.

chingor13:
> I see no calls to the synthtool library code at all. Did I miss one?

We have a step that runs synthtool templates if there is a `synth.py` that contains a `java.common_templates()` method call. It even parse out the list of excludes to preserve that functionality.

> If there are no calls to synthtool, then what do you think about creating a new repo just for java with this code and templates? Ah, well, that may make migration difficult because then there would be two copies of the templates to keep in sync.
> Anyway, migrating this code out of the synthtool repo would reduce your maintenance burden, so something to consider in the future.

I think that would be clean, but perhaps wait until after migrating to owlbot across the repos so the templates don't diverge.

PR #992, opened 2021-03-18
chore(python): add kokoro configs for periodic builds against head

tmatsuo:
This change should be non-destructive.

Note for library repo maintainers:
After applying this change, you can easily add (or change) periodic
builds against head by adding config files in google3.

See python-pubsub repo for example.


tmatsuo:
@dizcology FYI

busunkim96 (approved):
Thank you @tmatsuo!

PR #995, opened 2021-03-23
feat(bazel): Add clean_build option to synthtool

miraleung:
In `synth.py`, users will be able to enable build cleaning as follows. This is useful because there is otherwise no way to clear the Bazel cache via the shell.

This should temporarily affect only the PHP microgenerator. Its Bazel rules are implemented in such a way that clobbering the cache is needed if generator changes are made locally, but this will eventually be fixed in the microgenerator. Remote runs will not be affected.
```py
import synthtool.gcp as gcp

gapic = gcp.GAPICBazel()
gapic._clean_build = True
```


synthtool/gcp/gapic_bazel.py
```diff
@@ -47,6 +47,7 @@ def __init__(self):
         self._googleapis_private = None
         self._googleapis_discovery = None
         self._discovery_artifact_manager = None
+        self._clean_build = False
```
chingor13, line 50: Probably prefer this to be set via the constructor args (or keyword arg to `_generate_code()`). `_variables` are meant to signal private so you shouldn't configure this via `gapic._clean_build = True`.

chingor13 (changes requested):

chingor13:
If the `clean_build` option isn't meant to be left in `synth.py`, consider making it a utility method on GAPICBazel to explicitly clear the cache (rather than as a step in `_generate_code()`).

synthtool/gcp/gapic_bazel.py
miraleung: Done.

SurferJeffAtGoogle:
The code looks fine but I'm always hesitant to make changes that affect all languages.

In the synth.py files where we need to clean, could we instead call:
```.py
shell.run(["bazel", "clean", "--expunge", "--async"])
gapic = gcp.GAPICBazel()
gapic.php_library(..)
```

SurferJeffAtGoogle (changes requested):
See prior comment.

miraleung:
> The code looks fine but I'm always hesitant to make changes that affect all languages.
> 
> In the synth.py files where we need to clean, could we instead call:
> 
> ```python
> shell.run(["bazel", "clean", "--expunge", "--async"])
> gapic = gcp.GAPICBazel()
> gapic.php_library(..)
> ```

I'd tried this, but this throws an error since it's not executed in the Bazel workspace (which is accessible only inside `gcp.GAPICBazel()`(.

synthtool/gcp/gapic_bazel.py
```diff
@@ -182,6 +189,10 @@ def _generate_code(
         cwd = os.getcwd()
         os.chdir(str(api_definitions_repo))
 
+        if clean_build:
```
SurferJeffAtGoogle, line 192: I'd love to refactor it `_generate_code()`, turn it inside out, so there are 7 top-level functions calling common functions as needed, with no if statements.  But, I'm afraid of breaking everything.

Sorry you had to deal with this, Mira.

SurferJeffAtGoogle (approved):

SurferJeffAtGoogle:
I'll approve as soon as Kokoro tests pass.

chingor13 (approved):

PR #997, opened 2021-03-23
fix: node.owlbot_main() threw an exception

SurferJeffAtGoogle:
because it tried use a variable before it was assigned.  Added tests that execute all code paths through owlbot_main()


bcoe (approved):
I appreciate that there are now tests for this path 👍 

Only concern is that 295 files is _a lot_ of files, perhaps we could add two or three interesting files that exercise the code path?

SurferJeffAtGoogle:
295 is a lot. Do you see negative consequences to adding 295 files?  I see a small increase in the time to clone and checkout the repo.

bcoe:
> Do you see negative consequences to adding 295 files

For me it's mostly aesthetics when searching through the project, not a blocker.

PR #1013, opened 2021-03-31
test: add a golden test that runs the owlbot-java image on a test repo

chingor13:
As part of the cloud build process, after building the owlbot post-processor image, simulate running the post-processor against an input directory and directly compare against a set of golden output files.

### DIFF
--- a/docker/owlbot/java/bin/entrypoint.sh
+++ b/docker/owlbot/java/bin/entrypoint.sh
@@ -30,6 +30,11 @@ echo "Generating clirr-ignored-differences.xml..."
 /owlbot/bin/write_clirr_ignore.sh
 echo "...done"
 
+# fix license headers
+echo "Fixing missing license headers..."
+/owlbot/bin/fix_license_headers.sh
+echo "...done"
+
 # TODO: re-enable this once we resolve thrashing
 # restore license headers years
 # echo "Restoring copyright years..."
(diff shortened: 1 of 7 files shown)


docker/owlbot/java/cloudbuild.yaml
```diff
@@ -12,14 +12,26 @@
...
       ["test", "--image", "gcr.io/$PROJECT_ID/owlbot-java:$SHORT_SHA", "--config", "docker/owlbot/java/container_test.yaml"]
+    waitFor: ["build"]
+
+  # end-to-end tests
+  - name: "gcr.io/$PROJECT_ID/owlbot-java:$SHORT_SHA"
+    dir: docker/owlbot/java/tests/new-client/input
+    id: "new-client-run"
+    waitFor: ["build"]      
+  - name: "bash"
+    args: ["diff", "-rw", "input", "golden"]
```
bcoe, line 32: Will this exit `1` if the output from `new-client/input` differs from `golden`?

Perhaps add a note in the README for the container, regarding how to create a new golden fixture?

bcoe (approved):
This is slick 👍

docker/owlbot/java/cloudbuild.yaml
chingor13: > Will this exit `1` if the output from `new-client/input` differs from `golden`?

It does (maybe not with `1`, but definitely a non-zero exit code) and will cause the build to fail.

> Perhaps add a note in the README for the container, regarding how to create a new golden fixture?

Will do.

chingor13:
The presubmit will fail until #1009 is merged

PR #1015, opened 2021-03-31
fix(java): owlbot parent version should use existing parent version

### LINKED ISSUE #1014: Java: when adding new pom.xml files for new service versions, the wrong parent version is referenced

chingor13:
Fixes #1014


docker/owlbot/java/src/fix-poms.py
```diff
@@ -286,8 +286,8 @@ def main():
         existing_modules[parent_artifact_id] = module.Module(
             group_id=group_id,
             artifact_id=parent_artifact_id,
-            version="0.0.1-SNAPSHOT",
-            release_version="0.0.0",
+            version=main_module.version,
```
bcoe, line 289: When owl-bot ran was it just using the hard coded `0.0.1-SNAPSHOT` vs., whatever the library it was running against?

docker/owlbot/java/tests/new-version/golden/google-cloud-foo-bom/pom.xml
```diff
@@ -0,0 +1,95 @@
+<?xml version='1.0' encoding='UTF-8'?>
+<project xmlns="http://maven.apache.org/POM/4.0.0" xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance" xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>com.google.cloud</groupId>
+  <artifactId>google-cloud-foo-bom</artifactId>
+  <version>0.1.1-SNAPSHOT</version><!-- {x-version-update:google-cloud-foo:current} -->
```
bcoe, line 6: Is this the value that would have been incorrect.

docker/owlbot/java/cloudbuild_test.yaml
```diff
@@ -32,3 +32,11 @@ steps:
     args: ["diff", "-rw", "input", "golden"]
     dir: docker/owlbot/java/tests/new-client
     waitFor: ["new-client-run"]
+  - name: "gcr.io/$PROJECT_ID/owlbot-java:$SHORT_SHA"
+    dir: docker/owlbot/java/tests/new-version/input
```
bcoe, line 36: I like how test cases end up being additional steps on Cloud Build.

docker/owlbot/java/tests/new-version/golden/google-cloud-foo-bom/pom.xml
chingor13: Yeah, you can see the isolated failing check on the merge commit.

chingor13: From the log:
```
--- input/proto-google-cloud-foo-v2/pom.xml
+++ golden/proto-google-cloud-foo-v2/pom.xml
@@ -10,7 +10,7 @@
   <parent>
     <groupId>com.google.cloud</groupId>
     <artifactId>google-cloud-foo-parent</artifactId>
-    <version>0.0.1-SNAPSHOT</version><!-- {x-version-update:google-cloud-foo:current} -->
+    <version>0.1.1-SNAPSHOT</version><!-- {x-version-update:google-cloud-foo:current} -->
   </parent>
   <dependencies>
     <dependency>
```

docker/owlbot/java/src/fix-poms.py
chingor13: It's loading the list of modules/versions from the `versions.txt` file which does not contain the parent artifact versions (we assume it matches the main `google-cloud-X` artifact)

bcoe (approved):

PR #1016, opened 2021-04-01
fix: non-root users can run the nodejs docker post processor

SurferJeffAtGoogle:
Fixes https://github.com/googleapis/repo-automation-bots/issues/1585 via 3 methods:
1. Install node using apt-get instead of nvm.
2. Pre-install all the dependencies of protobufjs.  It's sneaky and tries to install things when you run it.
3. Make sure the /synthtool directory tree is readable by all.


bcoe (approved):
This looks good to me, I'll test in the morning.

bcoe:
@SurferJeffAtGoogle I feel like I'm missing something, I'm trying to run the container locally, I:

1. `cd ~/google/nodejs-secret-manager` (_which is where I have secret manager on my laptop_).
2. `docker run  --user $(id -u):$(id -g) --rm -v $(pwd):/repo -w /repo nodejs-testing`, where nodejs-testing is the testing repo I've created.
3. things seem to run as expected, but the README is not updated (_if I delete it, or change the `.repo-metadata.json`_).

PR #1017, opened 2021-04-02
feat(nodejs): templates_only can be specified to skip generating index.ts

bcoe:
I noticed that https://github.com/googleapis/nodejs-local-auth is failing with the error:

```
Traceback (most recent call last):
  File "/usr/local/lib/python3.9/runpy.py", line 197, in _run_module_as_main
    return _run_code(code, main_globals, None,
  File "/usr/local/lib/python3.9/runpy.py", line 87, in _run_code
    exec(code, run_globals)
  File "/synthtool/synthtool/languages/node.py", line 319, in <module>
    owlbot_main()
  File "/synthtool/synthtool/languages/node.py", line 286, in owlbot_main
    default_version = json.load(open(".repo-metadata.json", "rt"))["default_version"]
KeyError: 'default_version'
```

This is because OwlBot was added to a repo that does not have generated code (_just so we could take advantage of the shared templates._)

This is behavior I think is worth supporting, so that we can retire synthtool entirely.

### DIFF
--- a/synthtool/languages/node.py
+++ b/synthtool/languages/node.py
@@ -283,17 +283,19 @@ def owlbot_main(template_path: Optional[Path] = None):
     """
     logging.basicConfig(level=logging.DEBUG)
     # Load the default version defined in .repo-metadata.json.
-    default_version = json.load(open(".repo-metadata.json", "rt")).get("default_version", None)
-    templates_only = json.load(open(".repo-metadata.json", "rt")).get("templates_only", None)
+    default_version = json.load(open(".repo-metadata.json", "rt")).get(
+        "default_version", None
+    )
+    templates_only = json.load(open(".repo-metadata.json", "rt")).get(
+        "templates_only", None
+    )
     staging = Path("owl-bot-staging")
     s_copy = transforms.move
     if templates_only:
         # Some Node.js libraries use OwlBut purely for post-processing, see
         # https://github.com/googleapis/nodejs-local-auth
         common_templates = gcp.CommonTemplates(template_path)
-        templates = common_templates.node_library(
-            source_location="build/src"
-        )
+        templates = common_templates.node_library(source_location="build/src")
         s_copy([templates], excludes=[])
     else:
         if staging.is_dir():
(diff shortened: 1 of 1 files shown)


sofisl (approved):

synthtool/languages/node.py
```diff
@@ -283,36 +283,52 @@ def owlbot_main(template_path: Optional[Path] = None):
     """
     logging.basicConfig(level=logging.DEBUG)
     # Load the default version defined in .repo-metadata.json.
-    default_version = json.load(open(".repo-metadata.json", "rt"))["default_version"]
+    default_version = json.load(open(".repo-metadata.json", "rt")).get(
+        "default_version", None
+    )
+    templates_only = json.load(open(".repo-metadata.json", "rt")).get(
```
SurferJeffAtGoogle, line 289: We could instead inspect .OwlBot.yaml to see if there are any copies, and avoid adding another field to .repo-metadata.json.

synthtool/languages/node.py
```diff
@@ -283,36 +283,52 @@ def owlbot_main(template_path: Optional[Path] = None):
...
-        for version in versions:
-            library = staging / version
-            _tracked_paths.add(library)
-            s_copy([library], excludes=["README.md", "package.json", "src/index.ts"])
-        # The staging directory should never be merged into the main branch.
-        shutil.rmtree(staging)
+    if templates_only:
+        # Some Node.js libraries use OwlBut purely for post-processing, see
+        # https://github.com/googleapis/nodejs-local-auth
+        common_templates = gcp.CommonTemplates(template_path)
```
SurferJeffAtGoogle, line 297: I find it hard to trace the path through this function with the nested if statements.  We could reduce the nested if depth by one by structuring it like this:

```py
if templates_only:
  versions = []
  default_version = None
elif staging.is_dir():
  ...
else:
  ...

common_templates = gcp.CommonTemplates(template_path)
...
```

SurferJeffAtGoogle (changes requested):
The path through this function becomes hard to follow.  I'd rather leave a simple owlbot.py behind than

PR #1018, opened 2021-04-02
test(python): remove tests/unit from coverage

busunkim96:
Coverage should be measured against library code in `google/cloud`.


plamut (commented):
I'm not sure if we want that? One useful thing is that the coverage can warn you if a test is not executed, e.g. because of a typo in the test method name.

What would be the reasons to exclude the tests from coverage check?

busunkim96:
@plamut Ah that's fair. I was getting weird coverage drops from parameterized tests that were executed but not getting picked up in the coverage tests.

Closing.

PR #1021, opened 2021-04-05
chore: Add license headers for python config files

parthea:


synthtool/gcp/templates/python_library/docs/conf.py.j2
```diff
@@ -1,4 +1,17 @@
 # -*- coding: utf-8 -*-
+# Copyright 2021 Google LLC
+#
```
busunkim96, line 3: Did you check with OSPO that this is OK?

busunkim96 (approved):

PR #1024, opened 2021-04-05
feat: expose get_staging_dirs() for Python

SurferJeffAtGoogle:
Almost every synth.py for a Python library has naughty code.  So there's no hope of Python repos without owlbot.py files.  Instead, we'll replace the call to gapic.py_library() with a call to get_staging_dirs().


SurferJeffAtGoogle:
fyi @parthea

bcoe (approved):

PR #1026, opened 2021-04-06
chore: add deprecation warning template to readme

Neenu1995:


chingor13 (changes requested):
Add these to the existing `.readme-partials.yaml` - it's explicitly for the README and has key/value pairs for entries in the README template

PR #1029, opened 2021-04-07
feat: implement collect_versions_from_src()

SurferJeffAtGoogle:
`owlbot.py`s will need to call this function.

### DIFF
--- a/synthtool/languages/node.py
+++ b/synthtool/languages/node.py
@@ -24,6 +24,7 @@ from synthtool.sources import git
 from typing import Any, Dict, List, Optional
 import logging
 import shutil
+import pathlib
 
 _REQUIRED_FIELDS = ["name", "repository"]
 _TOOLS_DIRECTORY = "/synthtool"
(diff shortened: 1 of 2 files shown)


synthtool/languages/node.py
```diff
@@ -258,6 +259,20 @@ def postprocess_gapic_library_hermetic(hide_output=False):
     logger.debug("Post-processing completed")
 
 
+def collect_versions_from_src(default_version: str) -> List[str]:
```
bcoe, line 262: This is close to the same behave as `def detect_versions(path="./src") -> List[str]:`, which was added in an earlier iteration of the processor to solve this problem.

Can we remove `detect_version`?

bcoe (changes requested):

synthtool/languages/node.py
SurferJeffAtGoogle: Actually, I'll fix detect_version instead.().

PR #1031, opened 2021-04-08
fix: Add fetch for default branch 

billyjacobson:
Add fetch for default branch so autogen files can use the correct default branch during migration

### DIFF
--- a/synthtool/gcp/common.py
+++ b/synthtool/gcp/common.py
@@ -19,6 +19,7 @@ import shutil
 from copy import deepcopy
 from pathlib import Path
 from typing import Dict, List, Optional
+import requests
 import jinja2
 
 from synthtool import shell, _tracked_paths
@@ -57,6 +58,11 @@ class CommonTemplates:
                 self.excludes.append("samples/README.md")
 
         t = templates.TemplateGroup(self._template_root / directory, self.excludes)
+
+        repo = kwargs["metadata"]["repository"]
+        github_req = requests.get(f"https://api.github.com/repos/{repo}")
+        github_req.raise_for_status()
+        kwargs["metadata"]["repo"]["default_branch"] = github_req.json()["default_branch"]
 
         # TODO: migrate to python.py once old sample gen is deprecated
         if directory == "python_samples":
@@ -353,3 +359,9 @@ def _load_repo_metadata(metadata_file: str = "./.repo-metadata.json") -> Dict:
         with open(metadata_file) as f:
             return json.load(f)
     return {}
+
+def _get_default_branch_name(package_name: str) -> str:
+    r = requests.get(f"https://pypi.org/pypi/{package_name}/json")
+    r.raise_for_status()
+
+    return r.json()["info"]["version"]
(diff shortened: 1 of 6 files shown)


synthtool/gcp/common.py
```diff
@@ -353,3 +359,9 @@ def _load_repo_metadata(metadata_file: str = "./.repo-metadata.json") -> Dict:
         with open(metadata_file) as f:
             return json.load(f)
     return {}
+
+def _get_default_branch_name(package_name: str) -> str:
```
bcoe, line 363: It looks like this might not be used yet?

synthtool/gcp/common.py
```diff
@@ -58,6 +59,11 @@ def _generic_library(self, directory: str, **kwargs) -> Path:
 
         t = templates.TemplateGroup(self._template_root / directory, self.excludes)
 
+        repo = kwargs["metadata"]["repository"]
```
bcoe, line 62: This looks reasonable to me, but I think it would be worth adding a test here:

https://github.com/googleapis/synthtool/blob/master/tests/test_templates.py

bcoe (changes requested):
This is looking pretty reasonable to me, with the caveat that I think it would be good to add a test.

I think you could use something like [requests-mock](https://pypi.org/project/requests-mock/) so that you don't need to actually care about the traffic to GitHub.

synthtool/gcp/common.py
```diff
@@ -58,6 +59,11 @@ def _generic_library(self, directory: str, **kwargs) -> Path:
...
+        github_req = requests.get(f"https://api.github.com/repos/{repo}")
```
chingor13, line 63: @SurferJeffAtGoogle Will this be problematic for owl-bot (as it's not hermetic)?

synthtool/gcp/common.py
billyjacobson: fixed

synthtool/gcp/common.py
billyjacobson: added a test

synthtool/gcp/common.py
billyjacobson: I think Jeff is out til Friday? Is there anyone else we can ask about this? Or do you have thoughts on this?

bcoe: To me this seems like it's probably okay, as it will be applied in the post-processor, and will stay consistent for the repo itself?

bcoe (approved):

bcoe:
@billyjacobson thank you for adding this 👍 sorry it took a while to land, the beginning of the week was a bit swamped.

tests/test_common.py
```diff
@@ -41,6 +43,14 @@ def test_handles_empty_string():
     assert decamelize("") == ""
 
 
+def test_get_default_branch():
+    with requests_mock.Mocker() as m:
+        m.get(
```
bcoe, line 48: I really appreciate that you added a test for this :ok:

SurferJeffAtGoogle:
Sorry I didn't chime in on this sooner.

This is causing @sofisl and I problems because we're trying to generate templated code for a new library, that doesn't have a corresponding github repo yet.

With this PR, `_generic_library()`  fetches data from the network _during template generation_.  That's not what I imagined it would ever do.  That also means it will fail whenever the network is flaky.  It's very fragile.

@bcoe and @billyjacobson Can we observe or encode the default branch without making network requests?

PR #1032, opened 2021-04-08
feat: add excludes and patch callback arguments to owlbot_main

SurferJeffAtGoogle:
Should greatly simplify migrations.

Used the word "excludes" instead of "ignores" because existing synth.py files already use the term "excludes", so using the word "ignores" would make migrating more tedious, error-prone, and confusing.

Example of a migrated repo:
https://github.com/googleapis/nodejs-ai-platform/compare/owl-bot?expand=1

Also fixed a bug where `generate_index_ts()` was called even when `"src/index.ts"` was in the exclude list.

### DIFF
--- a/tests/test_node.py
+++ b/tests/test_node.py
@@ -12,16 +12,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import filecmp
+import os
+import pathlib
+import shutil
+import tempfile
+from pathlib import Path
 from unittest import TestCase
 from unittest.mock import patch
-import os
-from pathlib import Path
-from synthtool.languages import node
-import pathlib
-import filecmp
+
 import pytest
-import tempfile
-import shutil
+
+import synthtool as s
+from synthtool.languages import node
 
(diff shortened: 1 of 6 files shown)


tests/test_node.py
```diff
@@ -213,28 +217,104 @@ def test_owlbot_main(hermetic_mock):
...
+    ).read()
+    node.owlbot_main(TEMPLATES, templates_excludes=["src/index.ts"])
+    # confirm index.ts was overwritten by staging index.ts.
+    text = open("./src/index.ts", "rt").read()
+    assert original_text == text
+
+
+@patch("synthtool.languages.node.postprocess_gapic_library_hermetic")
+def test_owlbot_main_with_staging_patch_staging(hermetic_mock, nodejs_dlp):
+    def patch(library: Path):
```
bcoe, line 293: this is clever, I like it. Are you thinking it will work for hermetic replacements?

SurferJeffAtGoogle: Yep.

bcoe (approved):

PR #1045, opened 2021-04-14
fix: get_staging_dirs default_version should be optional

chingor13:


bcoe (approved):
nit: seems like it would be worth a test.

bcoe (approved):

Issue #1055, opened 2021-04-22
remove codecov integration


elharo:
go/codecov-faq

chingor13:
This was done

PR #1056, opened 2021-04-22
build: remove codecov.io

elharo:
toward #1055 

This is a start. there might be more we need to remove.


chingor13 (changes requested):
https://github.com/googleapis/synthtool/blob/master/synthtool/gcp/templates/java_library/.kokoro/build.sh#L110-L113 is also important for kokoro, or it may fail a few builds because its necessary credentials are being revoked here

elharo:
Done. PTAL @chingor13

elharo (commented):
PTAL

bcoe:
@elharo @chingor13, we've removed the internal codecov bash script, making it a noop for the time being which references the internal bug We've also rolled out changes for Node.js already removing the Codecov action.

Could we make a smaller PR that just removes Codecov from the Java GitHub actions?

bcoe:
@elharo @chingor13, we've removed the internal codecov bash script, making it a noop for the time being which references the internal bug We've also rolled out changes for Node.js already removing the Codecov action.

Could we make a smaller PR that just removes Codecov from the Java GitHub actions?

bcoe:
Codecov.io has been removed from Node.js/Java GitHub actions, and from kokoro jobs that were reporting to codecov.io.

PR #1058, opened 2021-04-23
chore(python): remove noxfile.py check for GOOGLE_APPLICATION_CREDENTIALS

busunkim96:
For Python system tests, don't enforce that `GOOGLE_APPLICATION_CREDENTIALS` is set, since the majority of tests can be run with user credentials.

### DIFF
--- a/synthtool/gcp/templates/python_library/CONTRIBUTING.rst
+++ b/synthtool/gcp/templates/python_library/CONTRIBUTING.rst
@@ -160,21 +160,7 @@ Running System Tests
   auth settings and change some configuration in your project to
   run all the tests.
 
-- System tests will be run against an actual project and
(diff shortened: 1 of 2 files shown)


parthea (approved):

tseaver:
FWIW, using the SDK credentials to access GCP APIs is explicitly [deprecated / discouraged](https://github.com/googleapis/google-auth-library-python/blob/04c870a5bcc2c25433774a27280505d3413550e3/google/auth/_default.py#L60-L70) by the auth library.

busunkim96:
@tseaver Thank you for raising that. I believe it should be removed given current auth guidance. Opened https://github.com/googleapis/google-auth-library-python/issues/744

PR #1071, opened 2021-05-10
build(node): explicit update to google-gax

bcoe:
We run the post-processor at a specific version of google-gax.


bcoe:
We just put out a new container today for Node.js, so I'd like to hold off on landing this PR for a day or two.

PR #1089, opened 2021-05-18
build(python): use pinned sphinx version on docs

dandhlee:
Updating the noxfile config for Python client libraries to make use of pinned Sphinx version that is regulated through `requirements.txt` and Renovate bot so folks working with client libraries and submitting PRs aren't surprised by sudden breaks in the doc tests because of unexpected Sphinx version updates.

Currently it only updates the noxfile config, is there a Synthtool way to be able to create the file that spreads across the repositories? I don't think the file creation is the issue, but I think the version might be an issue and Synthool later complaining if we do end up upgrading Sphinx version than the one shown and specified in the Synthtool. Should this be the approach we take, or once this PR gets submitted and the PRs get spread to the client libraries should we update those to include the `requirements.txt` files and then submit the PR afterwards?


busunkim96:
> Currently it only updates the noxfile config, is there a Synthtool way to be able to create the file that spreads across the repositories? I don't think the file creation is the issue, but I think the version might be an issue and Synthool later complaining if we do end up upgrading Sphinx version than the one shown and specified in the Synthtool. Should this be the approach we take, or once this PR gets submitted and the PRs get spread to the client libraries should we update those to include the requirements.txt files and then submit the PR afterwards?

If you think different repos will need to have different versions of the docs dependencies, I vote for the requirements file option. Renovate will send PRs to update the versions of those requirements.

If you think it's more likely that the _exact same version_ will be used across the repos, I'd prefer that it be pinned it in the `session.install`. This adds predictability to the docs build (no surprise upgrades) and you can change the version fairly easily through this template. It has the small downside that Renovate won't send PRs to update when there are new releases, but I don't think it's super critical that we're using the latest version of the docs dependencies. :) 

```py
session.install('alabaster==0.7.12', 'recommonmark==0.7.1', ...)
```

dandhlee:
I agree that there's no need to always use the latest version of Sphinx. As well, I don't think there has to be a need for a different Sphinx version to be used for different client libraries. I'll update it to use a pinned Sphinx version in the noxfile, and update here when necessary.

dandhlee:
Moving forward to pin the version here, and then use Renovate bot on the sphinx plugin repo to receive updates and then make future updates on here when necessary.

busunkim96 (approved):
Thank you!

PR #1091, opened 2021-05-19
chore: minimize noise from build scripts

igorbernstein2:
This should limit the amount of scrolling client devs have to do when debugging a test failure


synthtool/gcp/templates/java_library/.kokoro/coerce_logs.sh
```diff
@@ -28,7 +28,6 @@ job=$(basename ${KOKORO_JOB_NAME})
 echo "coercing sponge logs..."
 for xml in `find . -name *-sponge_log.xml`
 do
-  echo "processing ${xml}"
```
SurferJeffAtGoogle, line 31: Are you sure examining the xml is never useful?

What is the cost of the current noise?

igorbernstein2: I think its only useful when trying to debug the kokoro infrastructure (ie, why are my test results not showing up), which is rare. It is completely useless when trying to figure out why a new client feature broke the build. 
Currently all builds have fail at end flag set, so to figure out which module broke, I need to scroll to the end of the build log for the summary. However the end of the file has a thousand lines of file paths that I dont care about, which I have to scroll up until I get to useful info. If we remove this echo my debug workflow is:

scroll to end
identify broken module
find in page for module name

Neenu1995 (approved):

PR #1092, opened 2021-05-19
chore: dump maven version along with java

igorbernstein2:


SurferJeffAtGoogle (approved):

chingor13 (approved):

chingor13:
Note that we are installing `mvn` in our Dockerfile here: https://github.com/googleapis/testing-infra-docker/blob/7675981189962dbb32fa4bd550d4637018c6dc78/java/java8/Dockerfile#L22-L26

PR #1099, opened 2021-05-25
feat: record post-processor-changes.txt for nodejs docker image

SurferJeffAtGoogle:


docker/owlbot/nodejs/cloudbuild.yaml
```diff
@@ -1,9 +1,17 @@
 steps:
+    # Write the latest commit message to a file in the workspace.
+    - name: 'gcr.io/cloud-builders/git'
+      entrypoint: '/bin/sh'
+      args:
+        - '-c'
+        - 'git log -1 --format="%s%n%n%b%nSource-Link: https://github.com/googleapis/synthtool/commit/%H" > post-processor-changes.txt'
```
bcoe, line 7: Am I understanding correctly that we'll use the last commit to the synthtool repository at the time when the container is created?

SurferJeffAtGoogle: Yes.

bcoe (approved):

PR #1100, opened 2021-05-26
build: add auto-approve to Node libraries

sofisl:


synthtool/gcp/templates/node_library/.github/CODEOWNERS
```diff
@@ -11,3 +11,6 @@
 {%- else %}
 *     @googleapis/yoshi-nodejs
```
bcoe, line 12: Thinking out load, are we going to need to add our auto-approve user to each sub-team that owns repos, e.g., if a telemetry team owns logging repos, does the auto-approve user need to be part of it.

synthtool/gcp/templates/node_library/.github/auto-approve.yml
```diff
@@ -0,0 +1,12 @@
+rules:
+- author: "release-please[bot]"
+  title: "^chore: release"
+  changedFiles:
+  - "package\\.json$"
+  - "CHANGELOG\\.md$"
+  maxFiles: 3
+- author: "gcf-owl-bot[bot]"
```
bcoe, line 8: I think we should do the `.OwlBot.lock` config in a separate PR. I'm personally comfortable with automating merging these files. But @chingor13 had some concerns we should talk through as a group.

bcoe (changes requested):
I'm excited to turn on auto-merging releases, I think it might be worth discussing our OwlBot lock file merging with @chingor13 and @SurferJeffAtGoogle.

synthtool/gcp/templates/node_library/.github/CODEOWNERS
sofisl: Yes, we do, but this is only if they want auto-approve and they have subteams. I will add this to the README as well.

JustinBeckwith (changes requested):
If we land this as a `feat:` it's going to trigger release PRs.  Let's relabel this as `chore:`?

bcoe (approved):

JustinBeckwith (approved):

PR #1105, opened 2021-05-31
chore(owlbot-java): only add modules with version v1 or v2 to parent pom

Neenu1995:
Java-gsuite-addons has an extra set of protos that other protos in the library depend on. These are added to the repo as a separate module `proto-google-apps-script-type-protos`. Only proto-google-cloud-gsuite-addons-v1 depends on this module. But owlbot is currently configured to add all modules to its parent pom. Since the parent do not depend on this extra protos module, it fails the dependency check.
image

To solve this issue, the post-processor is being configured to add only the modules with a version (v1 or v2) in it.

### DIFF
--- a/docker/owlbot/java/src/fix-poms.py
+++ b/docker/owlbot/java/src/fix-poms.py
@@ -129,23 +129,24 @@ def update_cloud_pom(
     # insert proto dependencies after protobuf-java
     for m in proto_modules:
         if m.artifact_id not in existing_dependencies:
-            print(f"adding new dependency {m.artifact_id}")
-            new_dependency = etree.Element(
-                "{http://maven.apache.org/POM/4.0.0}dependency"
-            )
-            new_dependency.tail = "\n    "
-            new_dependency.text = "\n      "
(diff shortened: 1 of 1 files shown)


docker/owlbot/java/src/fix-poms.py
```diff
@@ -129,23 +129,24 @@ def update_cloud_pom(
...
-            new_group.tail = "\n      "
-            new_artifact = etree.Element(
-                "{http://maven.apache.org/POM/4.0.0}artifactId"
-            )
-            new_artifact.text = m.artifact_id
-            new_artifact.tail = "\n    "
-            new_dependency.append(new_group)
-            new_dependency.append(new_artifact)
-            dependencies.insert(proto_index + 1, new_dependency)
+            if "v1" in m.artifact_id or "v2" in m.artifact_id:
```
chingor13, line 132: Let's make this more forward-compatible. We do have v3 services and presumably will continue to have more

chingor13 (changes requested):

chingor13 (approved):

PR #1107, opened 2021-06-03
feat: add sub_directory argument to get_staging_dirs() 

parthea:
I had some trouble figuring out how to use `get_staging_dirs()` added in #1024 in repos that have multiple APIs in a single client. 

One solution is to add a `sub_directory` argument to [get_staging_dirs()](https://github.com/googleapis/synthtool/blob/master/synthtool/transforms.py#L280) so that only directories listed under the specified `sub_directory` are returned.

### DIFF
--- a/synthtool/transforms.py
+++ b/synthtool/transforms.py
@@ -294,7 +294,7 @@ def get_staging_dirs(
     """
 
     staging = Path("owl-bot-staging")
-    
+
     if sub_directory:
         staging /= sub_directory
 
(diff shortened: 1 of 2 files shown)


parthea:
I also need this feature for 3 other repos:
[python-firestore](https://github.com/googleapis/python-spanner)
[python-workflows](https://github.com/googleapis/python-workflows)
[python-bigtable](https://github.com/googleapis/python-bigtable)

parthea:
After discussion with @SurferJeffAtGoogle , the plan is to solve the problem directly in the `owlbot.py` of each repo instead adding this feature.

PR #1119, opened 2021-06-16
chore: drop 'u"' prefixes in 'docs/conf.py'

tseaver:


busunkim96 (approved):
@parthea Is the Dockerfile still configured to re-build on every change?

parthea:
@busunkim96 yes , PRs should open automatically when there are changes to python templates in this repo. The setting that controls that is in the trigger section of cloud build.

plamut (approved):

Issue #1118, opened 2021-06-16
[Python] Remove Python 2.7 from templates


anguillanneuf:
It's probably safe to remove `2.7` support for all libraries now?

https://github.com/googleapis/synthtool/blob/bb5ef5a99df6d439c63fddf0f7b6fe58e0679b10/synthtool/gcp/templates/python_samples/noxfile.py.j2#L91

PR #1127, opened 2021-06-18
docs: omit mention of Python 2.7 in 'CONTRIBUTING.rst'

### LINKED ISSUE #1126: Drop mention of Python 2.7 in generated docs for 'python_library'
We should no longer be generating docs which state / imply that our libraries support Python 2.7:  those which must still do so (e.g., `python-storage`) should not be copying `CONTRIBUTING.rst` in blindly.

tseaver:
Closes #1126


tseaver:
See also: #1118

parthea (approved):

PR #1141, opened 2021-06-26
chore: python 3.9 as the default python version for noxfile.py

parthea:
#1069 was merged which upgrades the python version in the post processor image from python 3.8 to 3.9. The default version for `noxfile.py` should also be updated.

### DIFF
--- a/synthtool/gcp/common.py
+++ b/synthtool/gcp/common.py
@@ -222,7 +222,7 @@ class CommonTemplates:
 
         # Set default Python versions for noxfile.py
         if "default_python_version" not in kwargs:
-            kwargs["default_python_version"] = "3.8"
+            kwargs["default_python_version"] = "3.9"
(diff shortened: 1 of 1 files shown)


parthea:
@busunkim96 Please could you review/approve?

busunkim96:
I'm curious if this is worth the trouble at this point in time? I know you've moved many repos to the templated noxfile but this will cause issues for any noxfiles that aren't - the noxfile will want to use 3.8 for `nox -s blacken` while the owlbot postprocessor will only have 3.9.

Perhaps it would be better to move the blackening outside of owlbot.py, so changing this version is not as much of a chore.

parthea:
Superseded by #1144 .

I've added an agenda item to the python client libraries meeting to discuss moving the `blacken` session outside of `owlbot.py`

PR #1155, opened 2021-07-14
build(python): remove python 3.7 from kokoro Dockerfile

parthea:
A lot of python repositories have a failing `Kokoro docs-presubmit` check with error `gpg: keyserver receive failed: No name`. For example, https://github.com/googleapis/python-asset/pull/227

The issue is that the `fetch_gpg_keys.sh` script is failing in the Docker file [here](https://github.com/googleapis/synthtool/blob/master/synthtool/gcp/templates/python_library/.kokoro/docker/docs/Dockerfile#L68) because the key [here](https://github.com/googleapis/synthtool/blob/master/synthtool/gcp/templates/python_library/.kokoro/docker/docs/fetch_gpg_keys.sh#L39) is no longer available on the keyserver.

This PR removes `fetch_gpg_keys.sh` as, in most repositories, we don't need to install additional python versions. We can use the default python version in the ubuntu 20.04 base image which is python 3.8.10.


plamut (approved):
Looks good, thanks! And just in the right time, I just notice the failing docs-presubmit tests myself, too.

busunkim96 (approved):

chingor13 (approved):

PR #1156, opened 2021-07-14
chore: add dependencyDashboardLabels to renovate.json

Neenu1995:


synthtool/gcp/templates/java_library/renovate.json
```diff
@@ -69,5 +69,8 @@
     }
   ],
   "semanticCommits": true,
-  "masterIssue": true
+  "masterIssue": true,
```
chingor13, line 72: Let's switch this to the non-deprecated config while we're at it
```suggestion
  "dependencyDashboard": true,
```

chingor13 (approved):

PR #1160, opened 2021-07-16
chore: add preserve_gapic to java library generation options

BenWhitehead:
There are cases when generating a library the gapic directory prefix should be preserved (veneer clients for example). This feature adds a new parameter `preserve_gapic` to `java.gapic_library` and `java.bazel_library`


chingor13 (approved):

chingor13:
Changed to a "chore" because this does trigger a new docker image build and the commit message would be a `feat`.

PR #1162, opened 2021-07-19
chore(java): add a method to copy a java method and rename it

Neenu1995:
chore: add method to deprecate java methods


synthtool/languages/java.py
```diff
@@ -551,3 +552,118 @@ class Example {
...
+
+                public String foo() {
+                    return "bar";
+                }
+            }
+
+        To deprecate the `main` method above, use:
+
+        DEPRECATION_WARNING = "This method is deprecated. Use {@link #{new_method}()} instead.
+        deprecate_method('path/to/file', 'public void main(String[] args)', DEPRECATION_WARNING.format(new_method="foo"))
```
chingor13, line 648: This doesn't match what the test is doing. The test has the user specify the leading open/close comment tags.

It would be nice if this detected any preceding javadoc and adds to the existing javadoc block, but if it's not possible, we should document the behavior with a caveat.

tests/testdata/SampleDeprecateMethodGolden.java
```diff
@@ -0,0 +1,38 @@
+/*
+ * Copyright 2020 Google LLC
```
chingor13, line 2:
```suggestion
 * Copyright 2021 Google LLC
```

chingor13 (changes requested):

tests/testdata/SampleCopyMethodGolden.java
```diff
@@ -0,0 +1,37 @@
+/*
+ * Copyright 2020 Google LLC
```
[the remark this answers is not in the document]
Neenu1995: The year is 2020 here because the SampleClass.java file has year 2020 and Assertion test will fail otherwise.

tests/testdata/SampleDeprecateMethodGolden.java
Neenu1995: Same here.

synthtool/languages/java.py
```diff
@@ -551,3 +552,118 @@ class Example {
...
             # print(line)
             fp.write(line)
+
+
+def copy_and_rename_method(filename: str, signature: str, before: str, after: str):
+    """Helper to make a copy an entire method and rename it.
+
+    Goes line-by-line to detect the start of the block. Determines
+    the end of the block by a closing brace at the same indentation
+    level. This requires the file to be correctly formatted.
```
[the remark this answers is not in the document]
Neenu1995: This duplicates the method body and rename it to a new name. Call to to both methods are separate.

synthtool/languages/java.py
Neenu1995: Done

tests/testdata/SampleDeprecateMethodGolden.java
```diff
@@ -0,0 +1,38 @@
...
+  public static void foo() {
+    System.out.println("bar");
+  }
+
+  /*
+   * @deprecated This method will be removed in the next major version.
+   * Use {@link #foo()} instead
+   */
+   @Deprecated
+     public static void foobar() {
```
chingor13, line 29: Is there a way to make this correct before having to run the formatter?

tests/test_language_java.py
```diff
@@ -120,6 +120,58 @@ def test_remove_method():
...
+    with tempfile.TemporaryDirectory() as tempdir:
+        shutil.copyfile(
+            "tests/testdata/SampleClass.java", tempdir + "/SampleClass.java"
+        )
+        DEPRECATION_WARNING = """  /*
+   * @deprecated This method will be removed in the next major version.
+   * Use {{@link #{new_method}()}} instead
+   */
+@Deprecated
+"""
```
chingor13, line 150: Usability wise, I would prefer if this parameter was just the deprecated javadoc message (`This method will be removed in the next major version....`. The `@Deprecated` annotation should be done by the method for me.

tests/testdata/SampleDeprecateMethodGolden.java
```diff
@@ -0,0 +1,39 @@
...
+  public static void foo() {
+    System.out.println("bar");
+  }
+
+  /*
+   * @deprecated This method will be removed in the next major version.
+   * Use {@link #foo()} instead
+   * foo has the same functionality as foobar.
+   */
+   @Deprecated
```
chingor13, line 29: This seems to have an extra leading space for some reason.

chingor13 (changes requested):

tests/testdata/SampleDeprecateMethodGolden.java
Neenu1995: Done

tests/test_language_java.py
Neenu1995: Done

tests/testdata/SampleDeprecateMethodGolden.java
Neenu1995: Fixed

tests/testdata/SampleDeprecateMethodGolden.java
```diff
@@ -0,0 +1,39 @@
...
+
+class ExampleClass {
+  public static void foo() {
+    System.out.println("bar");
+  }
+
+  /*
+   * @deprecated This method will be removed in the next major version.
+   * Use {@link #foo()} instead
+   * foo has the same functionality as foobar.
```
chingor13, line 27: Almost there - `@deprecated` javadoc should come after the main method description.

chingor13 (changes requested):

tests/testdata/SampleDeprecateMethodGolden.java
Neenu1995: done

tests/testdata/SampleDeprecateMethodGolden.java
```diff
@@ -0,0 +1,50 @@
...
+      System.out.println("this is a test " + bar);
+    }
+  }
+ /*
+  * @deprecated This method will be removed in the next major version.
+  * Use {@link #sample()} instead
+  */
+  @Beta
+@Generated()
+@Deprecated
```
chingor13, line 33: These are all incorrect, and we can fix this.

chingor13 (changes requested):

chingor13 (approved):

PR #1177, opened 2021-08-09
feat(autosynth)!: support alternate default branch

chingor13:
Can now force specifying the default branch via the `--default-branch` CLI argument or the `DEFAULT_BRANCH` environment variable. If not set, we default to detecting the default branch from the API.

All other inferred instances of default branch are now explicit and we require the caller to provide a value.

BREAKING CHANGE: methods where the default branch were optional are now require parameters


chingor13:
These tests are failing because the run on a docker image with `git` version 2.17.1. `git init -b <initial-branch>` was introduced in git 2.28.0.

chingor13:
Switched `git init -b <initial-branch>` to `git init` + `git checkout -b <initial-branch>`. It's not quite the same, but for tests it works.

autosynth/change_pusher.py
```diff
@@ -159,7 +167,7 @@ def push_changes(
             # Do a git dance to construct a branch with the commits squashed.
             temp_branch = str(uuid.uuid4())
             executor.check_call(["git", "branch", "-m", temp_branch])
-            executor.check_call(["git", "checkout", "master"])
+            executor.check_call(["git", "checkout", self.default_branch])
```
SurferJeffAtGoogle, line 170: We can leverage the `commit_count` to completely ignore the name of the default branch, like this:
https://github.com/googleapis/synthtool/compare/master...SurferJeffAtGoogle:no-master?expand=1#diff-14e3db0fd118f42dda8afc3bdf3d61f64370fc5c1a5ef34975d441337c413699

autosynth/providers/php.py
```diff
@@ -23,7 +23,7 @@ def list_repositories():
     # monorepo.
     repo = "googleapis/google-cloud-php"
 
-    tree = gh.get_tree(repo)
+    tree = gh.get_tree(repo, "master")
```
SurferJeffAtGoogle, line 26: This code will need to be revisited when they change their branch name to "main".

Can we save them the pain by calling `gh.get_default_branch()` here?

SurferJeffAtGoogle: Same story for the other languages too.

autosynth/synth.py
```diff
@@ -225,19 +225,23 @@ def _inner_main(temp_dir: str) -> int:
         "--branch-suffix", default=os.environ.get("BRANCH_SUFFIX", None)
     )
     parser.add_argument("--pr-title", default="")
+    parser.add_argument("--default-branch", default=os.environ.get("DEFAULT_BRANCH"))
```
SurferJeffAtGoogle, line 228: With my suggested change to `change_pusher.py`, I think this argument becomes unnecessary.

synthtool/sources/git.py
```diff
@@ -82,9 +86,9 @@ def clone(
             cmd = ["git", "clone", "--recurse-submodules", "--single-branch", url, dest]
             shell.run(cmd, check=True)
         else:
-            shell.run(["git", "checkout", "master"], cwd=str(dest), check=True)
+            shell.run(["git", "checkout", default_branch], cwd=str(dest), check=True)
```
SurferJeffAtGoogle, line 89: I think we could do language maintainers a favor by detecting the default branch here.  Otherwise they'll have to update code after they rename the branch.

I believe github renames the branch, so that after set the default branch to `main`, there is no more `master` branch.  Also, we clone the repo with `--single-branch` on line 86 so there should only be one branch.  Therefore, we could list the branches in cwd and see which branch name is present:
```
git branch --sort=committerdate
```

The `--sort=commiterdate` will break the tie if both names appear.  We'll take the one with the more recent commits.

SurferJeffAtGoogle (changes requested):
Thanks for digging into this.  Where possible, I'd rather avoid asking repo/language maintainers to specify a branch name.   It'll make migration easier.

autosynth/synth.py
chingor13: Done, we now always fetch the default branch. We still need the value so that we can send the pull request to the right base branch.

autosynth/change_pusher.py
chingor13: Done

synthtool/sources/git.py
chingor13: Added a helper function to detect the latest branch

synthtool/sources/git.py
```diff
@@ -44,6 +44,14 @@ def make_repo_clone_url(repo: str) -> str:
         return f"https://github.com/{repo}.git"
 
 
+def _local_default_branch(path: pathlib.Path) -> str:
```
SurferJeffAtGoogle, line 47: Add a doc string for this function.

synthtool/sources/git.py
```diff
@@ -44,6 +44,14 @@ def make_repo_clone_url(repo: str) -> str:
...
+    output = subprocess.check_output(
+        ["git", "branch", "--sort=committerdate", "--format=%(refname:short)"],
+        cwd=str(path),
+    ).decode("utf-8")
+    return output.splitlines()[-1]
```
SurferJeffAtGoogle, line 52: Will this return the correct result if the output of `git branch` is the following?

```
master
main
jeff-branch
```

SurferJeffAtGoogle (changes requested):

synthtool/sources/git.py
chingor13: I think I misunderstood your request for listing the local branches.

synthtool/sources/git.py
```diff
@@ -44,6 +44,33 @@ def make_repo_clone_url(repo: str) -> str:
...
+            ["git", "branch", "--sort=-committerdate", "--format=%(refname:short)"],
+            cwd=str(path),
+        )
+        .decode("utf-8")
+        .splitlines()
+    )
+    for branch in branches:
+        if branch == "master" or branch == "main":
+            return branch
+    return None
```
SurferJeffAtGoogle, line 71: `None` is not a `str` so I'm surprised that the linting tools permit this.

synthtool/sources/git.py
```diff
@@ -44,6 +44,33 @@ def make_repo_clone_url(repo: str) -> str:
...
+
+    Arguments:
+        path {pathlib.Path} - Path to the local git clone
+
+    Returns:
+        string -- The inferred default branch.
+    """
+    branches = (
+        subprocess.check_output(
+            ["git", "branch", "--sort=-committerdate", "--format=%(refname:short)"],
```
SurferJeffAtGoogle, line 62: Aha! That minus sign in front of commiterdate must mean descending order.  Cool.

SurferJeffAtGoogle (approved):

synthtool/sources/git.py
chingor13: Yep, much easier than iterating the list in reverse.

PR #1182, opened 2021-08-16
chore(java): update shared-config and shared-dependencies version in pom template

Neenu1995:


chingor13 (approved):

chingor13 (changes requested):
You'll need to update the golden files for the java image test

chingor13 (approved):

PR #1186, opened 2021-08-18
chore(python): refactor unit / system test dependency install

### LINKED ISSUE #1185: Python: refactor installing unit/system test dependencies
- [ ] Rather than using the template langage to generate `session.install` commands in `noxfile.py`, use simplers form to generate a standard set of variables at the top of the file.
- [ ] Break out installing dependencies in to separate helper functions using those variables, called by the `unit` / `system` sessions.
- [ ] Pass in the constraints explicitly to the helper functions, to allow for future sessions (e.g., passing `--pre` as a constraint, see https://github.com/googleapis/google-cloud-python/issues/10578).

tseaver:
Closes #1185.

Opened PR as a draft for discussion.

### DIFF
--- a/synthtool/gcp/templates/python_library/noxfile.py.j2
+++ b/synthtool/gcp/templates/python_library/noxfile.py.j2
@@ -48,7 +48,7 @@ UNIT_TEST_DEPENDENCIES = [{% for v in unit_test_dependencies %}
 UNIT_TEST_EXTRAS = [{% for v in unit_test_extras %}
     "{{v}}",{% endfor %}
 ]
-UNIT_TEST_EXTRAS_BY_PYTHON = { {% if unit_test_extras_by_python %}{% for python_version, extras in unit_test_extras_by_python.items() %}
+UNIT_TEST_EXTRAS_BY_PYTHON = {{ '{' }}{% if unit_test_extras_by_python %}{% for python_version, extras in unit_test_extras_by_python.items() %}
     "{{python_version}}": [{% for v in extras %}
         "{{v}}",{% endfor %}
     ],{% endfor %}{% endif %}
@@ -70,7 +70,7 @@ SYSTEM_TEST_DEPENDENCIES = [{% for v in system_test_dependencies %}
 SYSTEM_TEST_EXTRAS = [{% for v in system_test_extras %}
     "{{v}}",{% endfor %}
 ]
-SYSTEM_TEST_EXTRAS_BY_PYTHON = { {% if system_test_extras_by_python %}{% for python_version, extras in system_test_extras_by_python.items() %}
+SYSTEM_TEST_EXTRAS_BY_PYTHON = {{ '{' }}{% if system_test_extras_by_python %}{% for python_version, extras in system_test_extras_by_python.items() %}
     "{{python_version}}": [{% for v in extras %}
         "{{v}}",{% endfor %}
     ],{% endfor %}{% endif %}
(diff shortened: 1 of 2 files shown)


synthtool/gcp/templates/python_library/noxfile.py.j2
```diff
@@ -81,29 +126,38 @@ def lint_setup_py(session):
...
+    session.install(*UNIT_TEST_STANDARD_DEPENDENCIES, *constraints)
+
+    if UNIT_TEST_EXTERNAL_DEPENDENCIES:
+        session.install(*UNIT_TEST_EXTERNAL_DEPENDENCIES, *constraints)
+
+    if UNIT_TEST_LOCAL_DEPENDENCIES:
+        session.install("-e", *UNIT_TEST_LOCAL_DEPENDENCIES, *constraints)
+
+    if UNIT_TEST_DEPENDENCIES:
+        session.install("-e", *UNIT_TEST_DEPENDENCIES, *constraints)
```
tswast, line 139: These aren't supposed to be installed in "editable" mode, are they?

synthtool/gcp/templates/python_library/noxfile.py.j2
```diff
@@ -81,29 +126,38 @@ def lint_setup_py(session):
     session.run("python", "setup.py", "check", "--restructuredtext", "--strict")
 
 
+def install_unittest_dependencies(session, *constraints):
+    session.install(*UNIT_TEST_STANDARD_DEPENDENCIES, *constraints)
```
tswast, line 130: I'd prefer if we installed UNIT_TEST_STANDARD_DEPENDENCIES and UNIT_TEST_DEPENDENCIES all in one go.

synthtool/gcp/templates/python_library/noxfile.py.j2
```diff
@@ -23,13 +23,58 @@ import shutil
...
+
+UNIT_TEST_PYTHON_VERSIONS=[{% for v in unit_test_python_versions %}"{{v}}"{% if not loop.last %}, {% endif %}{% endfor %}]
+UNIT_TEST_STANDARD_DEPENDENCIES = [
+    "mock",
+    "asyncmock",
+    "pytest",
+    "pytest-cov",
+    "pytest-asyncio",
+]
+UNIT_TEST_EXTERNAL_DEPENDENCIES = [{% for v in unit_test_external_dependencies %}
```
tswast, line 39: Do we want to deprecate `unit_test_external_dependencies`? It seems quire redundant to me with `unit_test_dependencies`

synthtool/gcp/templates/python_library/noxfile.py.j2
tseaver: Likely not -- I doubt anybody is using a `git:` URL for a dependency.

synthtool/gcp/templates/python_library/noxfile.py.j2
tseaver: Hmm, I'm not sure I see why?  The prior template alwasy installed the "standard" deps (`"mock", "asyncmock", "pytest", "pytest-cov", "pytest-asyncio"`) in a separate step, prior to any others.

synthtool/gcp/templates/python_library/noxfile.py.j2
tseaver: 3ed5783

synthtool/gcp/templates/python_library/noxfile.py.j2
tseaver: d274ce3

synthtool/gcp/templates/python_library/noxfile.py.j2
tswast: The reason is that the pip resolver can run in one go. Otherwise subsequent pip installs might have to downgrade some packages.

PR #1189, opened 2021-08-23
ci: removing linkage-monitor from the required checks

suztomo:
Linkage Monitor has served to identify the dependency conflicts
between the source tree in a pull request and the latest Libraries
BOM. However the Libraries BOM now implements the process to sync
with the dependencies of the Google Cloud BOM and the Google Cloud
BOM achieves dependency convergence with regard to the shared
dependencies BOM. Therefore, Linkage Monitor does not give much
benefit to our library development process any more.
https://github.com/GoogleCloudPlatform/cloud-opensource-java/issues/2137

### DIFF
--- a/synthtool/gcp/templates/java_library/.github/sync-repo-settings.yaml
+++ b/synthtool/gcp/templates/java_library/.github/sync-repo-settings.yaml
@@ -31,7 +31,6 @@ branchProtectionRules:
   requiredStatusCheckContexts:
   - "dependencies (8)"
   - "dependencies (11)"
-  - "linkage-monitor"
   - "lint"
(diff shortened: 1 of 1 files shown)


suztomo:
```
nox > mypy synthtool autosynth
synthtool/log.py:21: error: Cannot assign to a type
synthtool/log.py:21: error: Incompatible types in assignment (expression has type "None", variable has type "Type[ColoredFormatter]")
synthtool/log.py:71: error: Incompatible types in assignment (expression has type "Formatter", variable has type "ColoredFormatter")
Found 3 errors in 1 file (checked 47 source files)
```

Retrying.

PR #1188, opened 2021-08-19
chore(python): rename default branch to main

parthea:
All googleapis python repos (with exception of tooling) have migrated to use a default branch of main. See tracking issue [here](https://github.com/googleapis/google-cloud-python/issues/10579).

PR #1198, opened 2021-08-30
chore(python): rename master_doc to root_doc

parthea:
Rename `master_doc` to `root_doc` in python templates. See sphinx docs:
https://www.sphinx-doc.org/en/master/usage/configuration.html#confval-root_doc

### DIFF
--- a/synthtool/gcp/templates/python_library/docs/conf.py.j2
+++ b/synthtool/gcp/templates/python_library/docs/conf.py.j2
@@ -76,8 +76,8 @@ source_suffix = [".rst", ".md"]
 # The encoding of source files.
 # source_encoding = 'utf-8-sig'
 
-# The master toctree document.
-master_doc = "index"
+# The root toctree document.
+root_doc = "index"
 
 # General information about the project.
 project = "{{ metadata['repo']['distribution_name'] }}"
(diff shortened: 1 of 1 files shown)


tswast:
Is Version 4 of Sphinx required for this? I recall we used to lock to an older Sphinx. I guess not anymore, seeing as https://github.com/googleapis/python-kms/pull/166 has a passing build?

dandhlee:
@tswast The old version was pinned because the old `sphinx-docfx-yaml` required sphinx version lower than 2 due to an incompatibility bug. With the new `gcp-sphinx-docfx-yaml` I don't think it's strictly required that it's Sphinx version 4 but allows us to use a much newer version.

parthea:
We've pinned to sphinx 4.0.1 in the templated noxfile [here](https://github.com/googleapis/synthtool/blob/master/synthtool/gcp/templates/python_library/noxfile.py.j2#L210). I'll keep an eye out for any failures before merging this change in downstream repos.

tswast (approved):

busunkim96 (approved):

tseaver:
@parthea for the sake of reducing bot scream, can we please do the replacement of `master` in templated files in a single PR (i.e., merge this change with PR #1188)?

PR #1114, opened 2021-06-14
chore(python): autogenerate docs/index.rst

parthea:
This PR is a first step towards auto generating `docs/index.rst` for *most* python repos. 

The function `detect_versions` in `synthtool/languages/python.py` is identical to the one used by `node` [here](https://github.com/googleapis/synthtool/blob/master/synthtool/languages/node.py#L217) so alternatively, we could consider moving the existing function `detect_versions` outside of `node.py`.

PR #1201, opened 2021-09-02
chore(python): add .github/CODEOWNERS as a templated file

parthea:
I tested these changes by running the following commands after cloning synthtool
```
git checkout template-codeowners
docker build -t testcodeowners -f docker/owlbot/python/Dockerfile .
```

Once the docker image was built, I ran the following command in the root of python-contact-center-insights
```
docker run --user $(id -u):$(id -g) --rm -v $(pwd):/repo -w /repo template-codeowners
```


tseaver:
@parthea Please do not merge this until checking that all "manual" libraries exclude `.github/CODEOWNERS` in their `owlbot.py`:  this PR would have negative impact on them, and make the discoverability / ease-of-customization for those repose unnecessarily harder.

parthea:
@tseaver ,  I agree. I've added this as a requirement.

busunkim96:
Was the check for `GAPIC_AUTO` removed intentionally?

tseaver:
@busunkim96
> Was the check for `GAPIC_AUTO` removed intentionally?

I think we need to exclude `.github/CODEOWNERS` from being copied altogeher for the manual repos.  They have local policies which no amount of templating / `s.replace`-ing will not be worse than just excluding.

parthea:
Capturing comment from https://github.com/googleapis/python-data-qna/pull/83 to be incorporated in this PR: 
> Could the synthtool template for CODEOWNERS note that the file is generated and that edits should be made to .repo-metadata.json?

synthtool/gcp/templates/python_library/.github/CODEOWNERS
```diff
@@ -0,0 +1,19 @@
+# Code owners file.
+# This file controls who is tagged for review for any given pull request.
+#
+# For syntax help see:
+# https://help.github.com/en/github/creating-cloning-and-archiving-repositories/about-code-owners#codeowners-syntax
+
+{% if 'codeowner_team' in metadata['repo'] %}
```
tswast, line 7: Looks like the BigQuery repos are missing this property. Are you planning on sending some PRs to update the repo metadata?

parthea: Yes, I also need to add `default_version` to `.repo-metadata.json` for #1114. I'm planning to add it at the same time. I see nodejs already uses `default_version` so I'm following their lead. https://github.com/googleapis/nodejs-iap/blob/main/.repo-metadata.json#L12

parthea: Done.

parthea:
@busunkim96 @tswast @tseaver This is ready for review

tswast (approved):

synthtool/gcp/templates/python_library/.github/CODEOWNERS
```diff
@@ -0,0 +1,18 @@
+# Code owners file.
```
tswast, line 1: Should probably have a comment indicating that one needs to edit the repo metadata and that this is templated.

tswast (changes requested):

synthtool/gcp/templates/python_library/.github/CODEOWNERS
parthea: Done

tswast (approved):

PR #1215, opened 2021-09-16
feat: introduce GraalVM testing build script changes

dzou:
This PR introduces the build script changes needed to introduce GraalVM testing to client library repositories. It is a follow-up to https://github.com/googleapis/java-shared-config/pull/314

As discussed in https://github.com/googleapis/java-shared-config/pull/314, the intention is to produce changes that are no-op's and gradually opt-in projects into GraalVM testing.

Summary of changes:

* **build.sh** - Add an additional switch case `graalvm` which contains the sequence of commands to install GraalVM and run unit and integration tests compiled with GraalVM native image. The switch case statement is only activated if the `JOB_TYPE` env variable is set to `graalvm` in the Kokoro `.cfg` file. Should not affect existing code which does not set `JOB_TYPE=graalvm`. 

* **graalvm-native.cfg** - Introduces the GraalVM `.cfg` that defines the Kokoro presubmit job.

In order to opt-in a project into GraalVM testing, we just need to add the corresponding `graalvm-native.cfg` in google3 for each repository to start running the new presubmit job for the specified repository.

Additional notes: @lesv suggested implementing a dedicated GraalVM container image for testing: https://github.com/googleapis/testing-infra-docker/issues/165

cc/ @lesv @Neenu1995 @suztomo

### DIFF
--- a/synthtool/gcp/templates/java_library/.kokoro/build.sh
+++ b/synthtool/gcp/templates/java_library/.kokoro/build.sh
@@ -69,6 +69,25 @@ integration)
       verify
     RETURN_CODE=$?
     ;;
+graalvm)
+    # Install GraalVM
+    graalvmDir=${KOKORO_ARTIFACTS_DIR}/graalvm
+    mkdir ${graalvmDir}
+    retry_with_backoff 3 10 \
+      curl --fail --show-error --silent --location \
+      https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-21.2.0/graalvm-ce-java11-linux-amd64-21.2.0.tar.gz \
+      | tar xz --directory ${graalvmDir} --strip-components=1
+
+    # Set GraalVM as the Java installation
+    export JAVA_HOME=${graalvmDir}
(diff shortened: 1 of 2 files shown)


synthtool/gcp/templates/java_library/.kokoro/build.sh
```diff
@@ -69,6 +69,25 @@ integration)
...
     RETURN_CODE=$?
     ;;
+graalvm)
+    # Install GraalVM
+    graalvmDir=${KOKORO_ARTIFACTS_DIR}/graalvm
+    mkdir ${graalvmDir}
+    retry_with_backoff 3 10 \
+      curl --fail --show-error --silent --location \
+      https://github.com/graalvm/graalvm-ce-builds/releases/download/vm-21.2.0/graalvm-ce-java11-linux-amd64-21.2.0.tar.gz \
+      | tar xz --directory ${graalvmDir} --strip-components=1
```
suztomo, line 79: Didn't the container get this?

PR #1236, opened 2021-10-06
chore: add php postprocessor

tmatsuo:
Also fixed a bug of merging behavior for binary files.

The files under `tests/fixtures/php` are for the integration test.

### DIFF
--- a/synthtool/languages/php.py
+++ b/synthtool/languages/php.py
@@ -89,9 +89,10 @@ def owlbot_copy_version(
     if copy_excludes is None:
         copy_excludes = DEFAULT_COPY_EXCLUDES
     # detect the version string for later use
-    entries = os.scandir(src / "src")
+    src_dir = src / "src"
+    entries = os.scandir(src_dir)
     if not entries:
-        logger.info("there is no src directory to copy")
+        logger.info("there is no src directory '%s' to copy", src_dir)
         return
     version_string = os.path.basename(os.path.basename(next(entries)))
     logger.debug("version_string detected: %s", version_string)


synthtool/languages/php.py
```diff
@@ -0,0 +1,184 @@
...
+
+STAGING_DIR = "owl-bot-staging"
+METADATA_DIR = "GPBMetadata"
+COPYRIGHT_REGEX = re.compile(r"Copyright (\d{4}) Google LLC$", flags=re.MULTILINE)
+OWLBOT_PY_FILENAME = "owlbot.py"
+DEFAULT_COPY_EXCLUDES: typing.List[str] = []
+
+
+@contextlib.contextmanager
+def pushd(d: Path):
```
SurferJeffAtGoogle, line 36: Cool.  I didn't know it was so easy to create a context with a generator.

synthtool/languages/php.py
```diff
@@ -0,0 +1,184 @@
...
+    """Copies files from a version subdirectory.
+    """
+    logger.debug("owlbot_copy_version called from %s to %s", src, dest)
+
+    if copy_excludes is None:
+        copy_excludes = DEFAULT_COPY_EXCLUDES
+    # detect the version string for later use
+    entries = os.scandir(src / "src")
+    if not entries:
+        logger.info("there is no src directory to copy")
```
SurferJeffAtGoogle, line 94: Include the directory that was inspected in the log message.  Something like `Found no src directory in ${src}`.

synthtool/languages/php.py
```diff
@@ -0,0 +1,184 @@
...
+    """
+    logger.debug("owlbot_patch called for %s", os.getcwd())
+
+    # Apply common replacements, currently nothing.
+    pass
+
+
+def owlbot_main(
+    src: Path, dest: Path, copy_excludes: typing.Optional[typing.List[str]] = None,
+) -> None:
```
SurferJeffAtGoogle, line 142: I expect the patch function to be passed as an argument here, defaulting to `owlbot_patch`.

synthtool/transforms.py
```diff
@@ -133,7 +133,15 @@ def _copy_dir_to_existing_dir(
                 os.makedirs(str(dest_dir), exist_ok=True)
                 source_path = Path(os.path.join(root, name))
                 if merge is not None and dest_path.is_file():
-                    _merge_file(source_path, dest_path, merge)
+                    try:
+                        _merge_file(source_path, dest_path, merge)
+                    except Exception as e:
+                        logger.info(
```
SurferJeffAtGoogle, line 139: `logger.exception(`

Then there's no need to pass e.

synthtool/transforms.py
```diff
@@ -133,7 +133,15 @@ def _copy_dir_to_existing_dir(
...
+                            "_merge_file failed for %s: %s, fall back to copy",
+                            source_path,
+                            e,
+                        )
+                        shutil.copy2(str(source_path), str(dest_path))
```
SurferJeffAtGoogle, line 144: I'm surprised we want to fall back to copy rather than fail completely.  Have you seen a case where it's better to fall back to copy?

synthtool/transforms.py
```diff
@@ -203,7 +211,13 @@ def move(
...
             # copy individual file
             if merge is not None and canonical_destination.is_file():
-                _merge_file(source, canonical_destination, merge)
+                try:
+                    _merge_file(source, canonical_destination, merge)
+                except Exception as e:
+                    logger.info(
+                        "_merge_file failed for %s: %s, fall back to copy", source, e
+                    )
+                    shutil.copy2(source, canonical_destination)
```
SurferJeffAtGoogle, line 220: I'm surprised we want to fall back to copy rather than fail completely. Have you seen a case where it's better to fall back to copy?

synthtool/languages/php.py
tmatsuo: Done (in the next commit)

synthtool/transforms.py
tmatsuo: PHP library has some binary files and synthtool dies when trying to `merge` those files.

tmatsuo:
@SurferJeffAtGoogle Ready for another look!

synthtool/transforms.py
SurferJeffAtGoogle: Ah, that's annoying.  To do this right would require running regex on the bytes.  But the signature for the _merge function requires strings, not bytes.  Ok.

synthtool/languages/php.py
SurferJeffAtGoogle: I don't see it yet.

PR #1262, opened 2021-10-21
chore(python):  Block pushing non-cloud libraries to Cloud RAD

parthea:
This PR will eliminate the need for the following replacements in owlbot.py:

```
# Block pushing non-cloud libraries to Cloud RAD
s.replace(
    ".kokoro/docs/common.cfg",
    r'value: "docs-staging-v2"',
    r'value: "docs-staging-v2-staging"'
)
```

As an example, see [this replacement](https://github.com/googleapis/python-analytics-admin/blob/main/owlbot.py#L47) in python-analytics-admin.


parthea:
@dandhlee Please could you take a look? This change will ensure that newly generated client libraries will automatically be configured correctly.

dandhlee (approved):
Yay! This looks great, thank you for doing this. It will be super helpful.

chingor13 (approved):

PR #1202, opened 2021-09-02
chore(java): install maven 3.8.1 at runtime

Neenu1995:
Gitbub Action updated LTS([18.04](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu1804-README.md), [20.04](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md)) versions of Github-Hosted runners for Ubuntu with the latest version of Apache Maven. The new version of Ubuntu has `Apache Maven 3.8.2` which is incompatible with `flatten-maven-plugin:1.2.7`. This caused all github actions using the latest Ubuntu version to fail.
```
Error:  Failed to execute goal org.codehaus.mojo:flatten-maven-plugin:1.2.7:flatten (flatten) on project google-cloud-bigquery: failed to create a clean pom: unable to create flattened dependencies: caught exception when flattening dependencies: Failed to read artifact descriptor for com.fasterxml.jackson.core:jackson-core::2.12.5: Could not find artifact com.fasterxml.jackson:jackson-base:pom:2.12.5 -> [Help 1]
```

The immediate solution is to use [this](https://github.com/marketplace/actions/setup-maven) community action to install maven 3.8.1 at runtime.

PR #1268, opened 2021-10-26
chore(java): remove pin on Apache Maven 3.8.1 from github actions

Neenu1995:
[Ubuntu 20.04.3 LTS](https://github.com/actions/virtual-environments/blob/main/images/linux/Ubuntu2004-README.md) has updated the maven version to 3.8.3 which resolves[ the issue](https://github.com/googleapis/synthtool/pull/1202) maven 3.8.2 had with flatten-maven-plugin


synthtool/gcp/templates/java_library/.github/workflows/ci.yaml
```diff
@@ -42,28 +34,20 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        java: [8, 11, 17]
+        java: [8, 11]
```
chanseokoh, line 37: 17 accidentally removed?

Neenu1995: Good catch. Fixed now.

PR #1270, opened 2021-11-01
feat: add sphinx-copybutton to generated docs

shivdhar:
Add [sphinx-copybutton](https://github.com/executablebooks/sphinx-copybutton) to generated docs to help easier copying of code snippets.

Note: I wasn't able to get synthtool to work for https://github.com/googleapis/python-vision:

```shell
➜  python-vision git:(main) synthtool              
2021-11-01 22:06:03,169 synthtool > /home/[REDACTED]/python-vision/synth.py not found.
NoneType: None                                                                                                                                                     /0.6s
```
Any help would be appreciated!

In the docs (https://github.com/googleapis/synthtool/blob/master/synthtool/README.md):
```
e.g. the synth.py for the Cloud Tasks API for Python, Java, Node.js, PHP, or Ruby.
```
the link `https://github.com/googleapis/python-tasks/blob/master/synth.py is broken.

### DIFF
--- a/synthtool/gcp/templates/python_library/noxfile.py.j2
+++ b/synthtool/gcp/templates/python_library/noxfile.py.j2
@@ -209,7 +209,7 @@ def docs(session):
     """Build the docs for this library."""
 
     session.install("-e", ".")
-    session.install("sphinx==4.0.1", "alabaster", "recommonmark")
+    session.install("sphinx==4.0.1", "alabaster", "recommonmark", "sphinx-copybutton", )
 
     shutil.rmtree(os.path.join("docs", "_build"), ignore_errors=True)
     session.run(
@@ -229,7 +229,7 @@ def docfx(session):
     """Build the docfx yaml files for this library."""
 
     session.install("-e", ".")
-    session.install("sphinx==4.0.1", "alabaster", "recommonmark", "gcp-sphinx-docfx-yaml")
+    session.install("sphinx==4.0.1", "alabaster", "recommonmark", "gcp-sphinx-docfx-yaml", "sphinx-copybutton")
 
     shutil.rmtree(os.path.join("docs", "_build"), ignore_errors=True)
     session.run(
@@ -246,7 +246,8 @@ def docfx(session):
             "sphinx.ext.napoleon,"
             "sphinx.ext.todo,"
             "sphinx.ext.viewcode,"
-            "recommonmark"
+            "recommonmark",
+            "sphinx_copybutton",
         ),
         "-b",
         "html",
--- a/synthtool/gcp/templates/python_library/noxfile.py.j2
+++ b/synthtool/gcp/templates/python_library/noxfile.py.j2
@@ -229,7 +229,7 @@ def docfx(session):
     """Build the docfx yaml files for this library."""
 
     session.install("-e", ".")
-    session.install("sphinx==4.0.1", "alabaster", "recommonmark", "gcp-sphinx-docfx-yaml", "sphinx-copybutton")
+    session.install("sphinx==4.0.1", "alabaster", "recommonmark", "gcp-sphinx-docfx-yaml")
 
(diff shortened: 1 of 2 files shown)


parthea (commented):
Hi @shivdhar,

Thanks for opening a PR! This looks like a really helpful feature however I'm hesitant to include another dependency in our docs.  I'd like to get more folks to review and provide feedback.

synthtool/gcp/templates/python_library/noxfile.py.j2
```diff
@@ -229,7 +229,7 @@ def docfx(session):
     """Build the docfx yaml files for this library."""
 
     session.install("-e", ".")
-    session.install("sphinx==4.0.1", "alabaster", "recommonmark", "gcp-sphinx-docfx-yaml")
+    session.install("sphinx==4.0.1", "alabaster", "recommonmark", "gcp-sphinx-docfx-yaml", "sphinx-copybutton")
```
tseaver, line 232: The `docfx` session generates YAML files, not HTML, so this change (and the next one) are unnecessary.

tseaver (commented):
I'm fine with the extra dependency here, given the DX improvement.  I don't think we can safely configure `copybutton_prompt_text = ">>> "` here, as our samples are not all (or even mostly) done in REPL style (likely only in some manual libraries, and even there not consistently).

synthtool/gcp/templates/python_library/noxfile.py.j2
dandhlee: As Tres mentioned, `docfx` generates YAML and thus we need additional support for this extra plugin in `https://github.com/googleapis/sphinx-docfx-yaml`. Rest assured, the YAML being generated using this plugin will have click-to-copy buttons enabled on code snippets.

As such, please revert this for `docfx` session.

dandhlee (dismissed):
I think it might be good to have for HTML outputs, but for `docfx` sessions it won't be necessary to have this enabled.

For the `docs` session, I'd like to see the Sphinx doc examples with this enabled to see before we make further changes.

synthtool/gcp/templates/python_library/noxfile.py.j2
shivdhar: Thanks for the helpful comments. I've reverted the changes for the`docfx` session.

shivdhar:
> For the docs session, I'd like to see the Sphinx doc examples with this enabled to see before we make further changes.

@dandhlee Absolutely agree here, however I was unable to run `synthtool` on `python-vision`, the library I was trying to configure it for. (Please see my [first comment](https://github.com/googleapis/synthtool/pull/1270#issue-770789323) in this thread.). Any tips would be appreciated!

dandhlee:
Hey @shivdhar, you won't need to run this for all repos (synthtool will update across 100+ repos), and for testing purpose that's not really necessary. This updates the noxfile (https://github.com/googleapis/python-compute/blob/main/noxfile.py, for example) with the content you updated, and running `nox -e docs` session through those repos will suffice.

We'll likely need to have an internal staging of the doc so we can take an actual look of what the new pages look like, I'm happy to volunteer if needed.

shivdhar:
@dandhlee  Sure, do let me know if I can help in any way!

tseaver:
@shivdhar WRT running `synthtool` in your `python-vision` repo:  you can pass the name of the script on the command line.  We use `owlbot.py` as the main synth-controlling script, so that would be:

```bash
$ /path/to/venv/bin/synthtool owlbot.py 
```

parthea:
I'm going to close this PR as docs have migrated to cloud.google.com which have a copy code sample button. For example, see https://cloud.google.com/python/docs/reference/accessapproval/latest/google.cloud.accessapproval_v1.services.access_approval.AccessApprovalClient

PR #1286, opened 2021-11-22
build: test node.js 16

JustinBeckwith:


synthtool/gcp/templates/node_library/.github/workflows/ci.yaml
```diff
@@ -9,10 +9,10 @@ jobs:
     runs-on: ubuntu-latest
     strategy:
       matrix:
-        node: [10, 12, 14, 15]
+        node: [10, 12, 14, 16, 17]
```
bcoe, line 12: Since 17 is only stable for 6 months, I'm tempted to just add `16`? I'm cautious of being somewhat sparing with our test matrix size, in case we do eventually start bumping into any quotas on GitHub's side.

JustinBeckwith: Entirely your call.  Traditionally I've liked having "latest", whatever that may be, in the matrix.  Invariably, someone runs with the hottest bits here, and if something breaks or causes trouble, we probably need to know about it.  That having been said, such occurrences are rare.  I'm cool either way, but for now I'll just remove.

parthea:
Assigning to @sofisl currently on rotation

bcoe:
we're going to start planning our migration off Node 10 next week, let's update our test matrix when this happens.

PR #1289, opened 2021-11-23
chore(java): allow owlbot to exclude modules in pom

Neenu1995:
Owlbot can now customize which dependencies and poms to ignore during post processing and add any extra modules to versions.txt. Three new fields have been added to the .repo_metadata.json file to make these customizations possible. 

- extra_versioned_modules
 Add the `extra_versioned_modules` field in .repo_metadata.json file to add that particular module to the versions.txt.This doesn't change the other owlbot behaviours in relation to dependency updates in pom files, in any way. It simply adds the extra module to versions.txt to be recognized by release please. 

- excluded_poms
 Add the `excluded poms` field in .repo_metadata.json file to avoid any pom files from post processing by owlbot. Owlbot ignores this pom while post processing, meaning, any new dependency that need to be added to that pom file will have to be added manually(Renovate will still update versions for existing dependencies).

- excluded_dependencies
 Add `excluded_dependencies` field to .repo_metadata.json file to ignore any dependencies while updating a pom file. This especially comes in handy for non-typical gapic libraries where there are missing bom, proto or gapic modules.

Example:
```
  "extra_versioned_modules": "google-cloud-foo-emulator",
  "excluded_poms": "google-cloud-foo-bom",
  "excluded_dependencies": "grpc-google-cloud-foo-v2,google-cloud-foo-bom"
```


docker/owlbot/java/src/fix-poms.py
```diff
@@ -297,6 +320,11 @@ def main():
         )
     parent_module = existing_modules[parent_artifact_id]
 
+    required_dependencies = existing_modules.copy()
+    for dependency_module in excluded_dependencies_list:
+        if dependency_module in required_dependencies:
+            del required_dependencies[dependency_module]
```
chingor13, line 326: Consider building a new instance additively vs. deleting from a copy.

chingor13 (changes requested):
This is pretty complex - let's add some more golden tests.

docker/owlbot/java/src/fix-poms.py
Neenu1995: Done

chingor13 (approved):
Please add some context to the PR body for posterity. We should also make sure to update any internal/external docs on these new fields

PR #1299, opened 2021-12-03
fix(java): Add -ntp flag to native image testing command

dzou:
This adds the -ntp flag to native image testing command to produce more readable test log outputs.


lesv (approved):

suztomo (approved):

chanseokoh (commented):
While we're at it, how about adding it to all other `mvn` commands? Like for test, lint, javadoc, etc. It's really annonying.

chanseokoh:
Also with `-B` where it's missing.

dzou:
> While we're at it, how about adding it to all other `mvn` commands? Like for test, lint, javadoc, etc. It's really annonying.

Yes, this can be done.

PR #1303, opened 2021-12-08
chore: update python-docs-samples link to main branch

nicain:
Hoping to get the same solution as https://github.com/googleapis/python-automl/pull/296 via an upstream fix, instead of modifying owlbot.py

CC @dandhlee


dandhlee:
Both links work fine, LGTM. Thank you for making this change!

dandhlee:
nit: the title used here is what owlbot will use to make changes throughout the repos, could you add a note that this is for updating the link for sample link? Something like `chore: update python-docs-samples link to main branch` would clear the confusion more.

parthea (approved):

PR #1304, opened 2021-12-09
chore: fix php post processor

tmatsuo:
Delete the staging directory


chingor13 (approved):

tmatsuo:
I think I understand what's happening. The kokoro test is running in trampoline. PHP test is running docker with mounting the test source files. The volume path needs to be in the host filesystem, so docker in docker can not see the files.

tmatsuo:
Yeah, it works

PR #1310, opened 2021-12-28
chore(nodejs): migrate README to use stable or preview release_levels

parthea:
This PR modifies the README template for nodejs to use `stable` or `preview` as required by go/library-data-integrity instead of `ga`, `beta`, `alpha`, `eap` or `deprecated`.


chingor13 (changes requested):
Does this need to accompany mass PRs to update the `.repo-metadata.json` files? Or is there another migration path we can take? (like map `alpha`, `beta`, `eap` to preview and `ga` to `stable`)

chingor13:
Another option (although maybe too late if it's already been designed) is to use a separate `stability` field rather than `release_level`.

bcoe:
I've landed https://github.com/googleapis/synthtool/pull/1312 as a migratory path towards the switch to `stable`/`preview`, this way we can land the update to `.repo-metadata.json` without having the README temoparily incorrect.

PR #1316, opened 2022-01-05
chore: use python-samples-reviewers

busunkim96:
Googlers, search for "Sample Reviews - Cloud Developer Relations" internally for a doc on policies around sample reviews.


dandhlee (approved):
@googleapis/python-samples-owners FYI.

PR #1319, opened 2022-01-10
chore(owlbot-java): disable re-writing clirr file in owlbot java post-processor

Neenu1995:


chingor13 (commented):
Probably want to fix the fixtures for tests

chingor13 (commented):
Note this will affect the initial codegen too. We will be missing the initial clirr-ignored-differences.xml files

PR #1322, opened 2022-01-10
test(nodejs): remove 15 add 16

bcoe:
Node 15 is EOL. Going forward I suggest that our matrix only reflect LTS versions.


ddelgrosso1 (approved):
Thanks Ben!

PR #1324, opened 2022-01-11
chore(python): fix undefined name 'glob' in samples noxfile

parthea:
PR https://github.com/googleapis/synthtool/pull/1321 added `glob.glob` but the import statement was missing.


plamut (approved):
Good that we caught this reasonably early

leahecole:
Oh NO!!! I am so sorry. I caught it in the template in p-d-s but not here. Thank you SO MUCH @parthea !!!!!

PR #1326, opened 2022-01-11
chore: add samples noxfile test

leahecole:
Add a test for the samples noxfile template that compiles the code, saves it to a temp file, and runs flake8 on it to both make sure there are no syntax errors and ensure there are no import errors

### DIFF
--- a/tests/test_sample_templates.py
+++ b/tests/test_sample_templates.py
@@ -47,7 +47,9 @@ def test_samples_billing():
     ).read_text()
     assert "and you will need to [enable billing]" in result
 
+
 # def test_samples_noxfile(template_kwargs, expected_text):
+
 
 def test_samples_noxfile():
     t = templates.Templates(PYTHON_SAMPLES)
--- a/tests/test_sample_templates.py
+++ b/tests/test_sample_templates.py
@@ -50,7 +50,7 @@ def test_samples_billing():
 # def test_samples_noxfile(template_kwargs, expected_text):
 
 def test_samples_noxfile(template_kwargs, expected_text):
-    t = templates.Templates(PYTHON_SAMPLES
+    t = templates.Templates(PYTHON_SAMPLES)
     result = t.render("noxfile.py.j2", **template_kwargs,).read_text()
     # Validate Python syntax.
     result_code = compile(result, "noxfile.py", "exec")
--- a/tests/test_sample_templates.py
+++ b/tests/test_sample_templates.py
@@ -12,6 +12,8 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+import tempfile
+
 from pathlib import Path
 
 from synthtool.sources import templates
@@ -49,12 +51,16 @@ def test_samples_billing():
 
 
 def test_samples_noxfile():
+    from flake8.checker import FileChecker
     t = templates.Templates(PYTHON_SAMPLES)
     result = t.render("noxfile.py.j2").read_text()
     # Validate Python syntax.
     result_code = compile(result, "noxfile.py", "exec")
-    # exec(result_code, {}, {})
     assert result_code is not None
+    result_code_temp_file = tempfile.TemporaryFile()
+    result_code_temp_file.write(result_code)
+    print(FileChecker(result_code_temp_file))
+    
 
 
 def test_samples_footer():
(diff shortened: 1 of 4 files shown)


tests/test_sample_templates.py
```diff
@@ -48,6 +48,19 @@ def test_samples_billing():
...
+# def test_samples_noxfile(template_kwargs, expected_text):
+
+
+def test_samples_noxfile():
+    t = templates.Templates(PYTHON_SAMPLES)
+    result = t.render("noxfile.py.j2").read_text()
+    # Validate Python syntax.
+    result_code = compile(result, "noxfile.py", "exec")
+    assert result_code is not None
+    # for expected in expected_text:
```
tswast, line 60: You could try calling [`exec(result_code, {}, {})`](https://docs.python.org/3.10/library/functions.html#exec). I believe that'll catch a little bit more, but it still won't catch missing imports if they're referenced within one of the session functions, as that code won't be executed, just the surrounding setup code.

tswast: Potentially we could write the results to a temp file and run `flake8` against it. https://flake8.pycqa.org/en/latest/internal/checker.html#flake8.checker.FileChecker

Doesn't look like it can take just a string.

tests/test_sample_templates.py
```diff
@@ -48,6 +50,19 @@ def test_samples_billing():
...
 
+def test_samples_noxfile():
+    from flake8.checker import FileChecker
+    t = templates.Templates(PYTHON_SAMPLES)
+    result = t.render("noxfile.py.j2").read_text()
+    # Validate Python syntax.
+    result_code = compile(result, "noxfile.py", "exec")
+    assert result_code is not None
+    result_code_temp_file = tempfile.TemporaryFile()
+    result_code_temp_file.write(result_code)
```
tswast, line 61: Probably should be `write(result)` here.

leahecole:
Hmm I'll debug this later today. It passes locally, but not here.

leahecole:
@tswast, do you have any good examples of using the `FileChecker` method you linked directly? I ran into Syntax errors doing that which is why I switched to using subprocess, which seemed simpler and worked locally, but is making things unhappy in Kokoro. I'm stuck as to trying to decide if I should try to figure out the syntax error rabbit hole or the Kokoro one.

tswast:
I do not. `subprocess` might be your best bet. [`flake8.main.cli.main`](https://flake8.pycqa.org/en/latest/internal/cli.html#api-documentation) looks promising, but might end up calling `sys.exit`, which isn't quite what we want.

tests/test_sample_templates.py
```diff
@@ -48,6 +51,20 @@ def test_samples_billing():
...
+    result = t.render("noxfile.py.j2").read_text()
+    # Validate Python syntax.
+    result_code = compile(result, "noxfile.py", "exec")
+    assert result_code is not None
+    # write rendered template to a file
+    result_code_temp_file = tempfile.NamedTemporaryFile()
+    result_code_temp_file.write(bytes(result, "utf-8"))
+    # run flake8, which will check for missing imports
+    # test fails if flake8 fails, no need to assert
+    subprocess.check_call(["flake8", result_code_temp_file.name])
```
tswast, line 65: I believe this is failing because there is no `PATH` environment variable. You can either try an absolute path here or copy the environment variables to the subprocess.

```
subprocess.check_call(..., env=os.environ.copy())
```

Reference:

https://stackoverflow.com/questions/5658622/python-subprocess-popen-environment-path
https://docs.python.org/3/library/subprocess.html#subprocess.Popen

tswast: Note: This still won't work on Windows https://bugs.python.org/issue15451

Not sure if we consider that a problem, though.

tswast: Probably best to ignore my first suggestion and use https://docs.python.org/3/library/shutil.html#shutil.which instead

subprocess.check_call([shutil.which("flake8"), ...])

tswast: I think you need to add `flake8` to the installed packages here: https://github.com/googleapis/synthtool/blob/69fda12e2994f0b595a397e8bb6e3e9f380524eb/noxfile.py#L39

leahecole: waiting to see if that works - if it does, does that mean it doesn't need to be in the requirements.txt?

tswast (approved):
Looks great!

bcoe (approved):

PR #1333, opened 2022-01-17
ci(python): run lint / unit tests / docs as GH actions

parthea:
I tested the changes using the following commands after cloning synthtool

```
git checkout add-lint-unittest-docs-checks-as-github-actions
docker build -f docker/owlbot/python/Dockerfile -t movetoghactions .
```

Once the docker image was built, I ran the following commands in the root of [python-access-approval](https://github.com/googleapis/python-access-approval).

```
docker run --user $(id -u):$(id -g) --rm -v $(pwd):/repo -w /repo movetoghactions
```

This should not impact the existing kokoro or docs pre-submit jobs as we are adding github action checks in parallel. Once this is merged, we can disable the kokoro/docs on a per repo basis.

PR #1335, opened 2022-01-20
chore(python): use cov_level in unittest gh action

parthea:
PR #1333 added [Github Actions](https://github.com/googleapis/synthtool/tree/master/synthtool/gcp/templates/python_library/.github/workflows) to python library templates. The coverage command [here](https://github.com/googleapis/synthtool/blob/master/synthtool/gcp/templates/python_library/.github/workflows/unittest.yml#L57) will fail on repos that don't have 100% coverage. This PR makes use of the `cov_level` template variable to set the coverage level to the correct value for each repo.


busunkim96 (approved):

parthea:
Wait for EOD to merge to avoid hitting kokoro github quota

Issue #1351, opened 2022-02-09
Add a link to `client library howto` to the README template


tmatsuo:
The README in our client libraries doesn't provide a document for how to customize client library's behavior (timeout, retry, etc). I suggest we add a link to the [Client Library Howto](https://github.com/googleapis/gax-nodejs/blob/main/client-libraries.md) in the READMEs (if that document is not permanent, we can use a specific commit hash).

Here's a draft:

```
This library is built on top of google-gax. To customize the common behavior of this client library (timeout, retry, etc), see [Client Library Howto](https://github.com/googleapis/gax-nodejs/blob/main/client-libraries.md).
```

bcoe:
I think setting the `timeout` on a single RPC call is a pretty common customization, and I often forget how to do it. Having a blurb along with an exmaple I think would be good, just a small snippet below the blurb.

bcoe:
I would put this directly below.

# Supported Node.js Versions

Perhaps in

# Advanced configuration

summer-ji-eng:
Modified the sentence a bit.
```
This library is built on top of google-gax as its transport layer. To customize the common behavior of this client library (timeout, retry, etc), see [Client Library Howto](https://github.com/googleapis/gax-nodejs/blob/main/client-libraries.md).
```

I suggest we could add in section 
Using the client library
...
[Insert here]
------------

Samples

PR #1352, opened 2022-02-09
fix: clarify the gax-nodejs usage in README

summer-ji-eng:
Address issue #1351


tmatsuo (approved):
LG, @bcoe might want to add a simple snippet.

bcoe (approved):
I think that this looks like a good addition to the docs, a good reason not to add a sample (which I'd originally suggested) is that the methods vary a bit on a client by client basis, and it might confuse a customer if we document a method that's not on the client.

A suggestion, since `timeout` is one of the more common settings we need to tweak for customers, perhaps we could add an example specifically for timeouts to:

https://github.com/googleapis/gax-nodejs/blob/main/client-libraries.md

parthea:
Assigning to @SurferJeffAtGoogle currently on rotation

bcoe:
@SurferJeffAtGoogle this will create 150 PRs when merged, so will leave to you whether you choose to merge this week or wait for me to do so during my rotation.

PR #1354, opened 2022-02-14
build: make OwlBot check required

chingor13:
Towards https://github.com/googleapis/repo-automation-bots/issues/2112

This file is only created upon initial generation. This will make it required for newly generated repositories.


Neenu1995 (approved):

chingor13:
This shouldn't actually propose any owlbot post-processor updates because it shouldn't touch the already created sync-repo-settings.yml configs.

PR #1358, opened 2022-02-25
chore(python): remove redundant samples

parthea:
After the auto generated samples were renamed in the python gapic generator in https://github.com/googleapis/gapic-generator-python/pull/1210, I'm having to manually remove the old samples as the [owlbot_main()](https://github.com/googleapis/synthtool/blob/master/synthtool/languages/python.py#L182) for python doesn't delete files. For example, see the manual changes in https://github.com/googleapis/python-analytics-admin/pull/190/commits/d5291355c220fe11617606b4068143bea009f767. This PR will remove the need to add manual commits when auto generated samples are renamed.


synthtool/languages/python.py
```diff
@@ -179,6 +182,9 @@ def owlbot_main() -> None:
 
     if default_version:
         for library in s.get_staging_dirs(default_version):
+            if clean_up_generated_samples:
+                shutil.rmtree("samples/generated_samples", ignore_errors=True)
+                clean_up_generated_samples = False
```
busunkim96, line 187: Why does this need to be set to `False` here?

synthtool/languages/python.py
```diff
@@ -55,7 +56,7 @@ def fix_pb2_headers(*, proto_root: str = "**/*_pb2.py") -> None:
     s.replace(
         proto_root,
         PB2_HEADER,
-        fr"\g<1>{LICENSE}\n\n\g<2>",  # change order to avoid stacking replacements
+        rf"\g<1>{LICENSE}\n\n\g<2>",  # change order to avoid stacking replacements
```
busunkim96, line 59: Does changing the order change behavior?

parthea: Possibly. I'll revert this change.

synthtool/languages/python.py
parthea: I set it to False after the directory `samples/generated_samples` is removed so that we don't remove the fresh samples.

parthea:
@busunkim96  Please could you review/approve?

chingor13 (approved):

Issue #1364, opened 2022-03-03
Python: Improve test performance by running tests in parallel


m-strzelczyk:
When working on code samples in Python language, I'm often irritated by how long it takes to run the full test suite for my PRs. I'd love to have support for `pytest-xdist` or `pytest-parallel` in the noxfiles that we force-sync in our Python repositories. All tests should always be written in a way, that they don't rely on any order and don't interfere with each other, so this should not cause any problems.

I'll provide a PR that supports those pytest extensions for review in a moment.

According to tests run on my local machine on the `python-compute` repository, the time it takes to run all tests for samples can be reduced to 5 minutes from almost half an hour IIRC.

parthea:
Closed via https://github.com/googleapis/synthtool/pull/1365

PR #1365, opened 2022-03-03
feat: Adding support for pytest-xdist and pytest-parallel

m-strzelczyk:
Closing #1364 

Auto-add parameters for `pytest` when we detect that `pytest-xdist` or `pytest-parallel` are installed through the `requirements.txt` or `requirements-test.txt` files.

Setting the options to auto, as we can't assume what's the best number would be for the various environments the tests will run on.


busunkim96 (approved):
Thank you @m-strzelczyk, this is an elegant way to speed up the samples test suite. Could you make the same change in https://github.com/GoogleCloudPlatform/python-docs-samples/blob/main/noxfile-template.py?

CC @googleapis/python-samples-reviewers.

PR #1369, opened 2022-03-16
chore(python): configure release-please on previous major versions

busunkim96:
Read the current Python library release version from the `setup.py` or `version.py` and add release-please branch configs for all previous major versions.

```yml
releaseType: python
handleGHRelease: true

# NOTE: this section is generated in owlbot_main()
# See https://github.com/googleapis/synthtool/blob/master/synthtool/languages/python.py
branches:
  - branch: v2
    handleGHRelease: true
    releaseType: python
  - branch: v1
    handleGHRelease: true
    releaseType: python
  - branch: v0
    handleGHRelease: true
    releaseType: python
```

Needs to be tested on more repositories.

### DIFF
--- a/synthtool/languages/python.py
+++ b/synthtool/languages/python.py
@@ -148,6 +148,42 @@ def py_samples(*, root: PathOrStr = None, skip_readmes: bool = False) -> None:
         s.copy([result], excludes=excludes)
 
 
+def configure_previous_major_version_branches() -> None:
+    """Allow releases from previous major version branches by editing
+     `.github/release-please.yml`.
+     
+     The current library version is obtained from `version.py` in `google/**/version.py`,
+     or the `setup.py`.
+
+     Releases are configured for all previous major versions. For example,
+     if the library version is currently 3.5.1, the release-please config
+     will include release branches for v2, v1, and v0.
+     """
+    if list(Path(".").glob("google/**/version.py")):
+        version_source_file = list(Path(".").glob("google/**/version.py"))[0].read_text()
+    else:
+        version_source_file = Path("setup.py").read_text()
+
+    # In version.py:    __version__ = "1.5.2"
+    # In setup.py:      version = "1.5.2"
+    VERSION_REGEX = '(?:__)?version(?:__)?\s?=\s?"(?P<major_version>\d)\.[\d\.]+"'
+    major_version = int(re.search(VERSION_REGEX, version_source_file).group("major_version"))
+
+    branch_release_config = """
+    # NOTE: this section is generated in owlbot_main()
+    # See https://github.com/googleapis/synthtool/blob/master/synthtool/languages/python.py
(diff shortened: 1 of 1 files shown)


parthea:
@busunkim96 , I applied the update on a repo that doesn't have a previous major version.  There are no branches listed under `branches:` . Can release-please handle an empty branches section or should we remove `branches:` if it's empty? 
```
releaseType: python
handleGHRelease: true
  # NOTE: this section is generated by synthtool.languages.python
  # See https://github.com/googleapis/synthtool/blob/master/synthtool/languages/python.py
  branches:
```

parthea (approved):
Added one observation, otherwise LGTM

tests/test_python_library.py
```diff
@@ -126,3 +128,37 @@ def test_split_system_tests():
...
+handleGHRelease: true
+  # NOTE: this section is generated by synthtool.languages.python
+  # See https://github.com/googleapis/synthtool/blob/master/synthtool/languages/python.py
+  branches:
+  - branch: v1
+    handleGHRelease: true
+    releaseType: python
+  - branch: v0
+    handleGHRelease: true
+    releaseType: python"""
```
chingor13, line 163: I think this indentation is wrong. `branches` should not have leading spaces

chingor13 (changes requested):

tests/test_python_library.py
busunkim96: Fixed!

synthtool/languages/python.py
```diff
@@ -148,6 +148,58 @@ def py_samples(*, root: PathOrStr = None, skip_readmes: bool = False) -> None:
...
+    if match is not None:
+        major_version = int(match.group("major_version"))
+    else:
+        raise RuntimeError(
+            "Unable to find library version in {} with regex {}".format(
+                version_file, VERSION_REGEX
+            )
+        )
+
+    if major_version > 0:
```
busunkim96, line 184: @parthea I added this to avoid generating an empty `branches:` section.

synthtool/languages/python.py
```diff
@@ -148,6 +148,58 @@ def py_samples(*, root: PathOrStr = None, skip_readmes: bool = False) -> None:
...
+  - branch: v{}
+    handleGHRelease: true
+    releaseType: python""".format(
+                v
+            )
+
+        release_please_yml = Path(".github/release-please.yml").read_text()
+        Path(".github/release-please.yml").write_text(
+            release_please_yml + branch_release_config
+        )
```
chingor13, line 200: Is this idempotent? It looks like it will just keep appending to the file.

Also, it might be cleaner to build it as a dict and export to yaml.

chingor13 (changes requested):

chingor13 (approved):

PR #1370, opened 2022-03-16
build(nodejs): remove staging fallback

bcoe:
I believe all Node.js libraries should be using a staging directory for generation, the fallback logic that remained led to a bad PR for gax:

https://github.com/googleapis/gax-nodejs/pull/1204/files

Because the directories in `src/` were seen as versions (_even though they are not_).

### DIFF
--- a/synthtool/languages/node.py
+++ b/synthtool/languages/node.py
@@ -317,13 +317,11 @@ def owlbot_main(
             s_copy([library], excludes=staging_excludes)
         # The staging directory should never be merged into the main branch.
         shutil.rmtree(staging)
-    else:
+    elif default_version:
         # Collect the subdirectories of the src directory.
         src = Path("src")
-        versions = [v.name for v in src.iterdir() if v.is_dir()]
         # Reorder the versions so the default version always comes last.
-        if default_version:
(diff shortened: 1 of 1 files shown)


chingor13 (approved):

synthtool/languages/node.py
```diff
@@ -317,14 +317,6 @@ def owlbot_main(
             s_copy([library], excludes=staging_excludes)
         # The staging directory should never be merged into the main branch.
         shutil.rmtree(staging)
-    else:
```
SurferJeffAtGoogle, line 320: Many pull requests for repos don't have staging directories at all, so I think we still need this code.

SurferJeffAtGoogle (changes requested):

synthtool/languages/node.py
```diff
@@ -317,13 +317,11 @@ def owlbot_main(
...
+    elif default_version:
```
bcoe, line 320: @SurferJeffAtGoogle if we want to keep the logic of grabbing versions from `src/`, I think this compromise would get us back to the existing behavior, while allowing `default_version` to be optional.

PR #1382, opened 2022-03-30
chore: Enable Size-Label bot in all googleapis NodeJs repositories

losalex:
Auto-label T-shirt size indicator should be assigned on every new pull request in all googleapis NodeJs repositories


synthtool/gcp/templates/node_library/.github/auto-label.yaml
```diff
@@ -0,0 +1,3 @@
+product: true
```
bcoe, line 1: did `product` default to true? wondering if we should just add requestsize, to make this change minimal.

losalex: > did `product` default to true? wondering if we should just add requestsize, to make this change minimal.

Actually you right - let me remove it

tmatsuo (approved):

PR #1461, opened 2022-06-16
docs: update how to run tests

yil532:
Update the README.md docs in synthtool.

### DIFF
--- a/synthtool/gcp/templates/node_library/samples/README.md
+++ b/synthtool/gcp/templates/node_library/samples/README.md
@@ -33,6 +33,16 @@ Before running the samples, make sure you've followed the steps outlined in
 
 `cd ..`
 
+## Run the test
+
+If you want to run all tests of the sample code, you can run the following commands from the root.
+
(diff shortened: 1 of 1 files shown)


synthtool/gcp/templates/node_library/samples/README.md
```diff
@@ -33,6 +33,16 @@ Before running the samples, make sure you've followed the steps outlined in
 
 `cd ..`
 
+## Run the test
+
+If you want to run all tests of the sample code, you can run the following commands from the root.
+
+`cd samples`
+
```
sofisl, line 41: I think you also need to link the local package by running `npm link ../`

We should also write a disclaimer that samples in the generated folder are autogenerated and not tested, thus not guaranteed to work without further tweaking. However, ones in the samples/ are and can be run from the samples directory.

sofisl (changes requested):

SurferJeffAtGoogle:
Closed due to inactivity.

PR #1470, opened 2022-06-28
feat: add nodejs mono repo post-processor

sofisl:


synthtool/gcp/common.py
```diff
@@ -326,6 +326,36 @@ def node_library(self, **kwargs) -> Path:
 
         return self._generic_library("node_library", **kwargs)
 
+    def node_mono_repo_library(self, relative_dir, **kwargs) -> Path:
```
sofisl, line 329: These changes are important

PR #1471, opened 2022-06-28
feat: add node mono repo post processor

sofisl:
Same as #1470, with easier-to-read diff


synthtool/gcp/common.py
```diff
@@ -326,6 +326,36 @@ def node_library(self, **kwargs) -> Path:
 
         return self._generic_library("node_library", **kwargs)
 
+    def node_mono_repo_library(self, relative_dir, **kwargs) -> Path:
+        # TODO: once we've migrated all Node.js repos to either having
+        #  .repo-metadata.json, or excluding README.md, we can remove this.
+        if not os.path.exists("./.repo-metadata.json"):
```
bcoe, line 332: I believe this check should include `relative_dir` too. I believe the main purpose of this logic is to exclude `samples/README.md` from templates, if there's no `samples/` directory.

synthtool/gcp/common.py
```diff
@@ -326,6 +326,36 @@ def node_library(self, **kwargs) -> Path:
 
         return self._generic_library("node_library", **kwargs)
 
+    def node_mono_repo_library(self, relative_dir, **kwargs) -> Path:
```
bcoe, line 329: since this logic is pretty much identical to `def node_library`, I might just make `node_library` take an optional `relative_dir` parameter, and then default to `./` unless there's an alternate `relative_dir` provided.

synthtool/languages/node_mono_repo.py
```diff
@@ -0,0 +1,397 @@
...
+            raise RuntimeError(
+                f"package.json is missing required fields {_REQUIRED_FIELDS}"
+            )
+
+        repo = git.parse_repo_url(data["repository"])
+
+        data["repository"] = f'{repo["owner"]}/{repo["name"]}'
+        data["repository_name"] = repo["name"]
+        data["lib_install_cmd"] = f'npm install {data["name"]}'
+        engines_field = re.search(r"([0-9][0-9])", data["engines"]["node"])
```
bcoe, line 54: I would switch this to `[0-9]+`, prevents a bug if Node ever goes to weekly releases like Chrome (which is up to version `101`).

synthtool/languages/node_mono_repo.py
```diff
@@ -0,0 +1,397 @@
...
+    Example:
+        @google-cloud/storage => google-cloud-storage-npm-token
+        dialogflow => dialogflow-npm-token
+
+    Args:
+        package: Name of the npm package.
+    Returns:
+        The name of the key to fetch the publish token.
+    """
+    return package_name.strip("@").replace("/", "-") + "-npm-token"
```
bcoe, line 109: We use a single token for all publication now, so this helper is obsolete, i.e., we should no longer be referencing `publish_token` anywhere.

synthtool/languages/node_mono_repo.py
```diff
@@ -0,0 +1,397 @@
...
+        filePath: the path of index.ts.
+    Returns:
+        Array of client name string extract from index.ts file.
+    """
+    with open(filePath, "r") as fh:
+        content = fh.read()
+    return re.findall(r"\{(.*Client)\}", content)
+
+
+def generate_index_ts(
```
bcoe, line 126: I think several of these helpers are identical to `languages/node.py`?

I think you could import the helpers from `node.py`, so that we don't have duplicate code (you can import individual functions).

synthtool/languages/node_mono_repo.py
```diff
@@ -0,0 +1,397 @@
...
+# 1. Copy all the staging files at the top level of our monorepo into our subfolder
+def walk_through_owlbot_dirs(dir: Path):
+    """
+    Walks through samples/generated to find all snippet metadata files, appends them to a list
+
+    Returns:
+    A list of all metadata files.
+    """
+    owlbot_dirs = []
+    for path_object in dir.glob("**/*"):
```
bcoe, line 273: I think there's a risk that this glob will get slow once we have dozens of repositories in the repo, as it's going to just create a list of all files in the entire mono-repo.

By convention, we're always going to have the `.OwlBot.yaml` in one location:

```
packages/
   a/.OwlBot.yaml
   b/.OwlBot.yaml
   b/.OwlBot.yaml
```

Given this, I would just use `os.listdir`, to list directories in `packages/`, and then check for a `.OwlBot.yaml` in each folder.

bcoe (changes requested):
Left some small refactoring suggestions. Mainly, I believe there are a few places where you could just import a method from `node.py`, rather than copying the helpers.

Algorithmically, I think there's a risk that glob won't scale well as we add 1000s of files. I would instead consider a simple directory list, since we know that files will be in a well known location.

Overall, I think this is great work 🎉 , it's great that you got through tests in place.

synthtool/gcp/common.py
sofisl: Normally, I think it would be better to simply modify the existing code rather than duplicating it. However, I think in this case it's better to duplicate it, because: 1) it makes it easier to read and debug (just follow the node_mono_repo naming), and 2) we are eventually migrating to a mono repo, and eventually all the split repo paths will be deprecated. Given that it's only a few helper methods that are truly being copied over, I'd opt for copying over adding a new path to a helper method (this goes for other helper methods that are similar as well). Furthermore, if there are more specific changes we need to make for mono-repos, it will be easier to add to a mono-repo only path rather than making sure it doesn't affect split repos.

bcoe (approved):

chingor13 (approved):

PR #1472, opened 2022-06-29
build: marking dependencies (8) non-required

suztomo:
Because our dependency tree do not changes between Java language versions, there's no point to require the dependency check each version. Keeping "dependencies (11)" intact.


blakeli0 (approved):

suztomo:
Merged this with "build:" prefix.

PR #1493, opened 2022-07-15
build: owlbot-java Dockerfile order change

suztomo:


suztomo:
The wait time is not that long. Keeping the current order.

PR #1496, opened 2022-07-15
fix(nodejs-mono-repo): get directory one path above, since owlbot.yaml is under .github/ now

sofisl:
Note: added fixture confirms correct behavior in tests.


sofisl:
See: https://github.com/googleapis/google-cloud-node/pull/3229

Shudogg (approved):

bcoe (approved):

PR #1505, opened 2022-07-22
feat(node-mono-repo): remove default_version from mono-repo post-processor

sofisl:
Pending conversation of whether we should remove default_version. See internal doc `Should Nodejs keep the concept of a default version?`


SurferJeffAtGoogle:
Closed due to inactivity.

PR #1512, opened 2022-07-28
feat(node-mono-repo): make years in node mono-repo templates variable

sofisl:


chingor13 (approved):
Need to fix the test assertion

sofisl:
@chingor13, I ended up making a new template so as to not create a bunch of PRs for our existing split repos.

PR #1524, opened 2022-08-08
fix(node-mono-repo): generate README and samples/README

sofisl:


synthtool/gcp/common.py
```diff
@@ -49,10 +49,10 @@ def __init__(self, template_path: Optional[Path] = None):
         self._templates = templates.Templates(self._template_root)
         self.excludes = []  # type: List[str]
 
-    def _generic_library(self, directory: str, **kwargs) -> Path:
+    def _generic_library(self, directory: str, relative_dir: str, **kwargs) -> Path:
```
bcoe, line 52: should `relative_dir` have a default value, it looks like it will be undefined in some cases?

bcoe (approved):

bcoe (commented):
Looks reasonable to me, but I notice a few failing tests, this may indicate that the default "" behavior for relative a realtive directory has issues.

synthtool/gcp/common.py
```diff
@@ -369,7 +371,7 @@ def render(self, template_name: str, **kwargs) -> Path:
         _tracked_paths.add(template)
         return template
 
-    def _load_generic_metadata(self, metadata: Dict):
+    def _load_generic_metadata(self, metadata: Dict, relative_dir=''):
```
bcoe, line 374: I'm not sure how the `fs` API works for Python, but if this is a directory path, it would make more sense to me to be `.`, or `./` -- if these values work.

bcoe (approved):

PR #1527, opened 2022-08-10
feat(node_mono_repo): copy the list sample to a quickstart

sofisl:
This feature will copy one of the generated samples to the quickstart and append the correct region tags, as long as it doesn't exist. This will ensure that when new libraries are generated, a quickstart is generated alongside it.


synthtool/languages/node_mono_repo.py
```diff
@@ -59,6 +59,44 @@ def read_metadata(relative_dir: str):
         return data
 
 
+def copy_sample_to_quickstart(relative_dir):
```
chingor13, line 62: nit: maybe `copy_list_sample_to_quickstart`. Without the `list` in the name, one might assume this is a general utility to copy a sample

synthtool/languages/common.py
```diff
@@ -34,7 +34,7 @@ def update_library_version(version: str, root_dir: str):
             f.truncate()
 
 
-def get_sample_metadata_files(dir: Path):
+def get_sample_metadata_files(dir: Path, regex=r"snippet_metadata"):
```
chingor13, line 37: If the script is using type-hints, please continue to annotate new arguments.

synthtool/languages/node_mono_repo.py
chingor13: Also, same comment on type-hints

tests/test_node_mono_repo.py
```diff
@@ -157,6 +157,18 @@ def test_write_release_please_config():
         )
 
 
+def test_copy_quickstart():
+    with util.copied_fixtures_dir(FIXTURES):
```
chingor13, line 161: The root `fixtures` directory contains folders that represent different repository structures. For this case, we should put the fixtures in a subfolder that represents the scenario you're testing (rather than in the fixtures root).

chingor13 (changes requested):

tests/test_node_mono_repo.py
sofisl: For this one, I'm following this pattern: https://github.com/googleapis/synthtool/blob/6d304d5520aee517aa8b54c2eb8bd1d9fcca8d0f/tests/test_node_mono_repo.py#L79

Otherwise, I'd need to create a whole new directory with the same fixtures, making the library bigger. What do you think?

chingor13: You are already essentially creating a new test scenario, it's just in the root `fixtures` directory.

chingor13 (approved):

bcoe (approved):

PR #1539, opened 2022-08-19
fix: hashes for python tools in nodejs libraries

SurferJeffAtGoogle:
One step towards fixing b/243042618.


SurferJeffAtGoogle:
@tmatsuo How do I test?

tmatsuo (approved):

PR #1579, opened 2022-09-13
fix(nodejs-mono-repo): fix broken links readme nodejs only searching for relative path

sofisl:


tests/test_node_mono_repo.py
```diff
@@ -88,13 +88,13 @@ def test_no_samples():
...
 
-def test_fix_sample_path():
+def test_fix_sample_file_path():
     with util.copied_fixtures_dir(FIXTURES):
         metadata = node_mono_repo.template_metadata(
             FIXTURES / "nodejs_mono_repo_with_samples" / "packages" / "datastore"
         )
         for sample in metadata["samples"]:
-            assert "/workspace/google-cloud-node/" not in sample["file"]
+            assert re.match("^packages/.*", sample["file"])
```
chingor13, line 97: I think this test is more readable and less fragile if this assertion is matching exactly what we expect. The input is fixed, so the expected output should also be explicitly declarable.

tests/test_node_mono_repo.py
```diff
@@ -88,13 +88,13 @@ def test_no_samples():
         assert len(metadata["samples"]) == 0
 
 
-def test_fix_sample_path():
+def test_fix_sample_file_path():
     with util.copied_fixtures_dir(FIXTURES):
```
chingor13, line 92: If you only need a subdirectory of the fixtures, you should specify the folder here. So maybe:
```suggestion
    with util.copied_fixtures_dir(FIXTURES / "nodejs_mono_repo_with_samples"):
```
will only copy that mono repo subfolder

chingor13 (changes requested):

chingor13 (approved):

PR #1584, opened 2022-09-19
docs: fix broken link in Node.js README

alexander-fenster:
The page on nodejs.org became 404, which unfortunately means we need to regenerate all READMEs.


alexander-fenster:
We'll wait for this to be fixed on nodejs.org instead.

PR #1586, opened 2022-09-21
ci: update to use hashed requirements.txt for pip

dandhlee:
Updating the rest of `pip install` found without using hashes. I think the directory can be just `.kokoro/requirements.txt` rather than `github/{{ metadata['repo']['repo'].split('/')[1] }}/.kokoro/requirements.txt`, confirmed it for `build.sh` but unsure about `test-samples-impl.sh` file. Please let me know if otherwise.

Although for testing purposes they do not need to be updated, wanted to make sure there's full coverage while we're working on this.

Updating GitHub Actions seemed a lot more harder to do this with hashes, and I think they'd be much harder for unexpected results to arise so going to leave them as is.

### DIFF
--- a/synthtool/gcp/templates/python_library/.kokoro/build.sh
+++ b/synthtool/gcp/templates/python_library/.kokoro/build.sh
@@ -36,7 +36,7 @@ export PROJECT_ID=$(cat "${KOKORO_GFILE_DIR}/project-id.json")
 # Remove old nox
 python3 -m pip uninstall --yes --quiet nox-automation
 
-# Install nox
+# Install dependencies
 python3 -m pip install --require-hashes -r .kokoro/requirements.txt
 python3 -m nox --version
 
--- a/synthtool/gcp/templates/python_library/.kokoro/build.sh
+++ b/synthtool/gcp/templates/python_library/.kokoro/build.sh
@@ -37,7 +37,7 @@ export PROJECT_ID=$(cat "${KOKORO_GFILE_DIR}/project-id.json")
 python3 -m pip uninstall --yes --quiet nox-automation
 
 # Install dependencies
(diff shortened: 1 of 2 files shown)


synthtool/gcp/templates/python_library/.kokoro/build.sh
```diff
@@ -36,8 +36,8 @@ export PROJECT_ID=$(cat "${KOKORO_GFILE_DIR}/project-id.json")
 # Remove old nox
 python3 -m pip uninstall --yes --quiet nox-automation
 
-# Install nox
-python3 -m pip install --upgrade --quiet nox
+# Install dependencies
+python3 -m pip install --require-hashes -r .kokoro/requirements.txt
```
dandhlee, line 40:
```suggestion
python3 -m pip install --require-hashes -r ./.kokoro/requirements.txt
```

synthtool/gcp/templates/python_library/.kokoro/test-samples-impl.sh
```diff
@@ -32,8 +32,8 @@ export PYTHONUNBUFFERED=1
 # Debug: show build environment
 env | grep KOKORO
 
-# Install nox
-python3.9 -m pip install --upgrade --quiet nox
+# Install dependencies
+python3.9 -m pip install --require-hashes -r .kokoro/requirements.txt
```
dandhlee, line 36:
```suggestion
python3.9 -m pip install --require-hashes -r ./.kokoro/requirements.txt
```

parthea (dismissed):

dandhlee:
Discussed with parthea@ offline. We'll revisit this idea after the migration that's planned.

PR #1596, opened 2022-09-29
chore: [java] stopping auto release

suztomo:
- chore: stopping auto-release
- ci: adding back deploy plugin to sample directory
- chore: removing unused readme.sh


meltsufin (approved):

synthtool/gcp/templates/java_library/samples/pom.xml
```diff
@@ -32,4 +32,25 @@
...
+
+  <build>
+    <plugins>
+      <plugin>
+        <groupId>org.apache.maven.plugins</groupId>
+        <artifactId>maven-deploy-plugin</artifactId>
+        <version>2.8.2</version>
+        <configuration>
+          <skip>true</skip>
+        </configuration>
```
suztomo, line 44: This is actually not needed but adding back here just in case we happen to have the sample directory part of a release. (The samples directory is not marked as a module in the root pom.xml, so it shouldn't happen.)

PR #1655, opened 2022-10-14
chore: add library version to snippet metadata

dizcology:
Update synthtool/languages/python.py to update generated snippet's metadata to include client library version.

Note that technically the version added here is only the most recent library release version, and not necessarily the expected version for the snippets (which may be generated for a new feature, not yet released and therefore without a library version).  A library's release version is known only at release time, and so release-please should be updating it: https://github.com/googleapis/google-cloud-python/pull/10677

Reference: https://github.com/googleapis/synthtool/pull/1356

### DIFF
--- a/synthtool/languages/python.py
+++ b/synthtool/languages/python.py
@@ -16,6 +16,7 @@ import re
 import sys
 
 import json
+from packaging import version as packaging_version
 from pathlib import Path
 import shutil
 from typing import Any, Dict, List
@@ -24,6 +25,7 @@ import yaml
 import synthtool as s
 from synthtool import _tracked_paths, log, shell
 from synthtool.gcp.common import CommonTemplates, detect_versions
+from synthtool.languages import common
 from synthtool.sources import templates
 
 PathOrStr = templates.PathOrStr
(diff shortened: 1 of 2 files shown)


SurferJeffAtGoogle (commented):
Looks good to me, but seek @parthea's approval.

synthtool/languages/python.py
```diff
@@ -160,41 +163,54 @@ def py_samples(*, root: PathOrStr = None, skip_readmes: bool = False) -> None:
...
     The current library version is obtained from `version.py` in `google/**/version.py`,
     or the `setup.py`.
 
-    Releases are configured for all previous major versions. For example,
-    if the library version is currently 3.5.1, the release-please config
-    will include v2, v1, and v0.
+    Raises:
+        RuntimeError: If no version string found.
     """
+    version_paths = list(Path(".").glob("google/**/version.py")) + [Path("setup.py")]
```
parthea, line 175: Please could you add `"google/**/gapic_version.py"` as well or change this to `list(Path(".").glob("google/**/*version.py"))`

parthea (commented):
I'm wondering if this change should be in [gapic-generator-python](https://github.com/googleapis/gapic-generator-python) instead of synthtool.

synthtool/languages/python.py
dizcology: Updated - please take a look.

dizcology:
Moving this change to gapic-generator-python makes a lot of sense.  Currently I don't think the library generator looks at the libraries' release version, so some more investigation will be needed.

dizcology:
We will add the `version` field in snippet metadata elsewhere instead.

PR #1691, opened 2022-11-01
chore: require hashes when installing dependencies in owlbot postprocessor job

mpeddada1:
Follow-up PR to https://github.com/googleapis/synthtool/pull/1592


mpeddada1:
Installing synthtool without deps results in: ` NameError: name 'ColoredFormatter' is not defined`

meltsufin (approved):

PR #1709, opened 2022-11-11
chore: update pre-commit config file

sadikkuzu:


dandhlee:
Hey @sadikkuzu, could you file an issue first to discuss why these changes are needed? Did you run into any problems?

sadikkuzu:
> Hey @sadikkuzu, could you file an issue first to discuss why these changes are needed? Did you run into any problems?

Hi @dandhlee 
Instead of a new issue, we can discuss here, ain't we?
Well, the problem is that the versions are not up to date. Do you still want to use them?
Thanks in advance.
All the best.

dandhlee:
Should a change not be necessary then a PR is also not necessary. Determining that is the first step usually.

I'm hesitant to do a major version bump as this change will roll out to 150+ packages. If there are issues you're running into as a result of lower versions, I'd be happy to look into those. Otherwise, incremental updates would be preferred or just not needed at all.

sadikkuzu:
> Should a change not be necessary then a PR is also not necessary. Determining that is the first step usually.
> 
> I'm hesitant to do a major version bump as this change will roll out to 150+ packages. If there are issues you're running into as a result of lower versions, I'd be happy to look into those. Otherwise, incremental updates would be preferred or just not needed at all.

Ok

PR #1711, opened 2022-11-15
chore(deps): update docfx session to require py3.9

dandhlee:
The new release for `gcp-sphinx-docfx-yaml` requires Python3.9. Pinning only for docfx session, this is not required for other jobs.


tmatsuo (approved):
BTW, the way nox installs dependencies is not checking the package hashes.

dandhlee:
Good point - I'll file an issue to convert them to use the hashed requirements file, perhaps separate for nox.

dandhlee:
Filed #1712, will follow up with the investigation.

PR #1715, opened 2022-11-22
build(node_mono_repo): run node post processor only on changed files

sofisl:
Changes the default behavior to search for files that have diverged from main.

Adds CLI args to use with the Docker container to specify running the post processor on some directories, or "all" directories


docker/owlbot/nodejs_mono_repo/Dockerfile
```diff
@@ -50,5 +53,4 @@ RUN find /synthtool -type d -exec chmod a+x {} \;
 RUN cd /synthtool && mkdir node_modules && npm i gts@3.1.0 google-gax@3.5.2 \
     typescript@4.7.4 @google-cloud/typeless-sample-bot@1.0.3
 
-ENTRYPOINT [ "/bin/bash" ]
-CMD [ "/synthtool/docker/owlbot/nodejs_mono_repo/entrypoint.sh" ]
+ENTRYPOINT [ "/bin/bash", "/synthtool/docker/owlbot/nodejs_mono_repo/entrypoint.sh" ]
```
bcoe, line 56: curious what the reason for switching from `CMD` to `ENTRYPOINT` was?

synthtool/languages/node_mono_repo.py
```diff
@@ -349,12 +351,32 @@ def walk_through_owlbot_dirs(dir: Path):
...
     packages_to_exclude = [r"gapic-node-templating", r"node_modules"]
+    if search_for_changed_files:
+        subprocess.run(["git", "fetch", "origin", "main", "--deepen=200"])
     for path_object in dir.glob("packages/**/.OwlBot.yaml"):
         if path_object.is_file() and not re.search(
             "(?:% s)" % "|".join(packages_to_exclude), str(Path(path_object))
         ):
-            owlbot_dirs.append(str(Path(path_object).parents[0]))
-
+            if search_for_changed_files:
```
bcoe, line 360: I would pull this into a helper method, like `collect_changed_files`, and run `suprocess.run` once to collect the full list of changed files, rather than calling it multiple times with `Path(path_object).parents[0]` as an argument.

synthtool/languages/node_mono_repo.py
```diff
@@ -349,12 +351,32 @@ def walk_through_owlbot_dirs(dir: Path):
...
+                            "origin/main...",
+                            Path(path_object).parents[0],
+                        ]
+                    ).returncode
+                    == 1
+                ):
+                    owlbot_dirs.append(str(Path(path_object).parents[0]))
+            else:
+                owlbot_dirs.append(str(Path(path_object).parents[0]))
+    for path_object in dir.glob("owl-bot-staging/*"):
```
bcoe, line 376: I like this approach, you don't even need to use the `git` call, you just need to check whether `owl-bot-staging` exists 😄

synthtool/languages/node_mono_repo.py
```diff
@@ -460,19 +475,43 @@ def owlbot_main(
...
     template_path: Optional[Path] = None,
     staging_excludes: Optional[List[str]] = None,
     templates_excludes: Optional[List[str]] = None,
     patch_staging: Callable[[Path], None] = _noop,
 ):
-    owlbot_dirs = walk_through_owlbot_dirs(Path.cwd())
-    for dir in owlbot_dirs:
-        owlbot_main(
-            dir, template_path, staging_excludes, templates_excludes, patch_staging
+    if specified_owlbot_dirs and specified_owlbot_dirs[0] == "all":
```
bcoe, line 484: nit: what if an empty list represented `all`, I think with this approach you could never have a directory called `all`?

tests/test_node_mono_repo.py
```diff
@@ -389,13 +394,40 @@ def test_owlbot_main_with_staging_ignore_index(hermetic_mock, nodejs_mono_repo):
...
     )
     # confirm index.ts was overwritten by staging index.ts.
     text = open("./packages/dlp/src/index.ts", "rt").read()
     assert original_text == text
 
 
+@patch("subprocess.run")
+def test_walk_through_owlbot_dirs_git_diff(mock_subproc_popen):
+    process_mock = Mock()
+    attrs = {"communicate.return_value": ("output", "error")}
```
bcoe, line 409: what is the output of `subprocess.run` in this example. I wasn't clear what this was testing, vs., `test_walk_through_owlbot_dirs`.

tests/test_node_mono_repo.py
```diff
@@ -427,4 +460,22 @@ def patch(library: Path):
...
         # just confirm it doesn't throw an exception.
-        node_mono_repo.owlbot_entrypoint(template_path=TEMPLATES)
+        node_mono_repo.owlbot_entrypoint(
+            template_path=TEMPLATES, specified_owlbot_dirs=["all"]
+        )
+
+
+def test_entrypoint_args_with_arg():
+    node_mono_repo.walk_through_owlbot_dirs = MagicMock()
+    node_mono_repo.owlbot_entrypoint(specified_owlbot_dirs=["all"])
```
bcoe, line 470: It would also be good to have a test case that takes a subset of `specified_owlbot_dirs`, and limits to this list.

bcoe (changes requested):
Left some review comments.

docker/owlbot/nodejs_mono_repo/Dockerfile
sofisl: This is so that we can pass command-line arguments to the Docker container, i.e.,:

`docker run ${IMAGE} packages/google-cloud-compute,packages/google-cloud-asset`

See example: https://github.com/googleapis/repo-automation-bots/blob/main/packages/owlbot-bootstrapper/common-container/Dockerfile which then passes these args: https://github.com/googleapis/repo-automation-bots/blob/e3c22cecfb1495b36f50bf5d35aebedc3e2696d1/packages/owlbot-bootstrapper/cli/cloudbuild-owlbot-bootstrapper.yaml#L24

synthtool/languages/node_mono_repo.py
sofisl: I considered that, but I think we want the default behavior to only run for changed files. In other words, normally, owlbot will just call the post processor without any args, like here: https://github.com/googleapis/repo-automation-bots/blob/32e02e7d8b497bb0a0a684c89615efab6ace75c8/packages/owl-bot/cloud-build/update-pr.yaml#L58

In that case, we only want it to check for changed files. So the reason I added an "all" argument was in case someone, for some reason, wanted to run it on all the subdirectories from the command line, i.e.:

```
docker run ${IMAGE} all
```

tests/test_node_mono_repo.py
sofisl: I am just checking that subprocess.run is called, since that is what checks git for any differences from main. If it was called, it means that we appropriately were going to check if there were any differences from main (which is what setting `search_for_changed_files` should do)

synthtool/languages/node_mono_repo.py
bcoe: I think I'd opt for `--all` rather than `all` (so that you don't have a weird edge case around folders named `all`). If this is annoying, since you now need to support flags, vs., just positional arguments, why don't we just add a comment:

```
# TODO: add support for running against all folders.
```

bcoe (approved):

PR #1716, opened 2022-11-24
fix: Pin flake8 to v < 6.0.0

Mariatta:
flake8 v 6.0.0 introduces backward incompatible change. flake8-import-order is broken because of it.
Until there is an updated version of flake8-import-order, we will need to pin flake8 to v<6.0.0


Mariatta:
A fix was made in flake8-import-order but they have not cut a release yet.
https://github.com/PyCQA/flake8-import-order/pull/190

Mariatta:
Closing in favor of #1717

PR #1724, opened 2022-12-03
build: have Kokoro grab service account credentials from secret that will be rotated

sofisl:
see cl/492588523

see https://github.com/googleapis/nodejs-speech/pull/979 as a test - samples are passing

Screen Shot 2022-12-06 at 2 56 16 PM

### DIFF
--- a/synthtool/gcp/templates/node_library/.kokoro/samples-test.sh
+++ b/synthtool/gcp/templates/node_library/.kokoro/samples-test.sh
@@ -19,7 +19,7 @@ set -eo pipefail
 export NPM_CONFIG_PREFIX=${HOME}/.npm-global
 
 # Setup service account credentials.
-export GOOGLE_APPLICATION_CREDENTIALS=${KOKORO_GFILE_DIR}/service-account.json
+export GOOGLE_APPLICATION_CREDENTIALS=${SECRET_LOCATION}/long-door-651-kokoro-system-test-service-account
 export GCLOUD_PROJECT={{ test_project or 'long-door-651' }}
 
 cd $(dirname $0)/..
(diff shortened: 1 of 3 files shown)


synthtool/gcp/templates/node_library/.kokoro/samples-test.sh
```diff
@@ -19,7 +19,7 @@ set -eo pipefail
 export NPM_CONFIG_PREFIX=${HOME}/.npm-global
 
 # Setup service account credentials.
-export GOOGLE_APPLICATION_CREDENTIALS=${KOKORO_GFILE_DIR}/service-account.json
+export GOOGLE_APPLICATION_CREDENTIALS=${SECRET_LOCATION}/long-door-651-kokoro-system-test-service-account
 export GCLOUD_PROJECT={{ test_project or 'long-door-651' }}
```
tmatsuo, line 23: I'm afraid that the env var `SECRET_LOCATION` is not available in the test shell script because we don't export it in `populate-secret.sh`. Maybe `${KOKORO_GFILE_DIR}/secret_manager/long-door-651-kokoro-system-test-service-account`?

sofisl: I think when the samples tests are run locally, Kokoro is not involved so we wouldn't need to be concerned with grabbing the secrets from secret manager. Normally, we'd just authenticate using the same instructions we provide customers (i.e., ADC)

bcoe (approved):

bcoe (commented):
Should this change be in our mono-repo configuration templates too?

tmatsuo (approved):

tmatsuo (approved):

PR #1739, opened 2022-12-19
build: have Kokoro grab service account credentials from secret that will be rotated for system tests

sofisl:
This should look exactly like: https://github.com/googleapis/synthtool/pull/1724


bcoe (approved):

timscharfenort8989 (commented):
Guz

PR #1741, opened 2022-12-21
chore: Upgrade to later version of gcp-uploader and protobuf

lqiu96:
Update to use the latest version of gcp-uploader package and unpin the dependencies.


mpeddada1:
@lqiu96 After this change, we will be able to re-enable renovate-bot for these files: https://github.com/googleapis/synthtool/blob/91c4e39e1c67f497441ca0e62f232424fe2af993/renovate.json#L5

lqiu96:
Seeing this error:
```
Step #9: ERROR: In --require-hashes mode, all requirements must have their versions pinned with ==. These do not:
Step #9:     SecretStorage>=3.2 from https://files.pythonhosted.org/packages/54/24/b4293291fa1dd830f353d2cb163295742fa87f179fcc8a20a306a81978b7/SecretStorage-3.3.3-py3-none-any.whl (from keyring==23.4.1->-r requirements.txt (line 223))
```

lqiu96:
> @lqiu96 After this change, we will be able to re-enable renovate-bot for these files:
> 
> https://github.com/googleapis/synthtool/blob/91c4e39e1c67f497441ca0e62f232424fe2af993/renovate.json#L5

@mpeddada1 Any objections if I include that change in this PR?

I think the change would be
```
"ignorePaths": ["*/**", "synthtool/gcp/templates/java_library/.kokoro/requirements.txt", "synthtool/docker/owlbot/java/src/requirements.txt"]
```
becomes
```
"ignorePaths": ["*/**"],
```
Is that right?

synthtool/gcp/templates/java_library/.kokoro/requirements.txt
```diff
@@ -434,23 +385,22 @@ six==1.16.0 \
...
     --hash=sha256:71c644c5369f4a6e07636f0aa966270449561fcea2e3d6747b8d23efaa9d7832 \
     --hash=sha256:9fe5ea21568a0a70e50f273397638d39b03353731e6cbbb3fd8502a33fec40bc
-    # via
-    #   -r requirements.in
-    #   importlib-metadata
+    # via importlib-metadata
+
+# WARNING: The following packages were not pinned, but pip requires them to be
+# pinned when the requirements file includes hashes. Consider using the --allow-unsafe flag.
+# setuptools
```
mpeddada1, line 406: Following the advice given here, how about we generate the requirements file using `--allow-unsafe`?

synthtool/gcp/templates/java_library/.kokoro/requirements.txt
```diff
@@ -434,23 +385,22 @@ six==1.16.0 \
...
     --hash=sha256:8abb2f1d86890a2dfb989f9a77cfcfd3e47c2a354b01111771326f8aa26e0254
     # via
-    #   -r requirements.in
     #   gcp-docuploader
     #   google-auth
     #   python-dateutil
-typing-extensions==4.1.1 \
-    --hash=sha256:1a9462dcc3347a79b1f1c0271fbe79e844580bb598bafa1ed208b94da3cdcd42 \
-    --hash=sha256:21c85e0fe4b9a155d0799430b0ad741cdce7e359660ccbd8b530613e8df88ce2
-    # via -r requirements.in
```
mpeddada1, line 444: Ah looks like this dependency is removed from the newly generated requirements.txt file. Could we verify if it is needed? Looking at the logs of an older kokoro `javadoc` or `stage` job (from Sept) may help us determine if it was installed.

synthtool/gcp/templates/java_library/.kokoro/requirements.txt
```diff
@@ -132,12 +125,11 @@ cryptography==38.0.3 \
     --hash=sha256:dfb4f4dd568de1b6af9f4cda334adf7d72cf5bc052516e1b2608b683375dd95c \
     --hash=sha256:ed7b00096790213e09eb11c97cc6e2b757f15f3d2f85833cd2d3ec3fe37c1722
     # via
-    #   -r requirements.in
     #   gcp-releasetool
     #   secretstorage
-gcp-docuploader==0.6.3 \
-    --hash=sha256:ba8c9d76b3bbac54b0311c503a373b00edc2dc02d6d54ea9507045adb8e870f7 \
-    --hash=sha256:c0f5aaa82ce1854a386197e4e359b120ad6d4e57ae2c812fce42219a3288026b
+gcp-docuploader==0.6.4 \
```
mpeddada1, line 130: Nice! Looks like we can finally upgrade gcp-docuploader:)

mpeddada1:
> @mpeddada1 Any objections if I include that change in this PR?
> 
> I think the change would be
> 
> ```
> "ignorePaths": ["*/**", "synthtool/gcp/templates/java_library/.kokoro/requirements.txt", "synthtool/docker/owlbot/java/src/requirements.txt"]
> ```
> 
> becomes
> 
> ```
> "ignorePaths": ["*/**"],
> ```
> 
> Is that right?

Making the change in this PR sounds good! And yup, it would become `["*/**"]`. Oh one note, could you also update `https://github.com/googleapis/synthtool/blob/master/docker/owlbot/java/src/requirements.in` before removing it from the ignorePaths option?

mpeddada1 (commented):
Thanks @lqiu96!

synthtool/gcp/templates/java_library/.kokoro/requirements.txt
lqiu96: Yeah I do see this in the job from ~Oct:
```
Collecting typing-extensions==4.1.1
  Downloading typing_extensions-4.1.1-py3-none-any.whl (26 kB)
Collecting urllib3==1.26.12
  Downloading urllib3-1.26.12-py2.py3-none-any.whl (140 kB)
Collecting zipp==3.6.0
  Downloading zipp-3.6.0-py3-none-any.whl (5.3 kB)
Using legacy 'setup.py install' for pyperclip, since package 'wheel' is not installed.
Installing collected packages: pyasn1, urllib3, six, rsa, pycparser, pyasn1-modules, protobuf, idna, charset-normalizer, certifi, cachetools, requests, googleapis-common-protos, google-auth, cffi, zipp, typing-extensions, jeepney, google-crc32c, google-api-core, cryptography, secretstorage, pyparsing, markupsafe, importlib-metadata, google-resumable-media, google-cloud-core, python-dateutil, pyperclip, pyjwt, packaging, keyring, jinja2, google-cloud-storage, colorlog, click, attrs, gcp-releasetool, gcp-docuploader
    Running setup.py install for pyperclip: started
    Running setup.py install for pyperclip: finished with status 'done'
```

Though I did copy this in from other repos so I need to find earlier versions.

synthtool/gcp/templates/java_library/.kokoro/requirements.txt
lqiu96: Yep, will do.

synthtool/gcp/templates/java_library/.kokoro/requirements.txt
lqiu96: I've added in the `typing-extensions` dependency.

lqiu96:
> Oh one note, could you also update `https://github.com/googleapis/synthtool/blob/master/docker/owlbot/java/src/requirements.in` before removing it from the ignorePaths option?

@mpeddada1 What did you mean by update? Should I just unpin the dependencies or pin the current latest versions or something else?

mpeddada1:
> @mpeddada1 What did you mean by update? Should I just unpin the dependencies or pin the current latest versions or something else?

I think we can now update the dependencies in the associated [requirement.txt](https://github.com/googleapis/synthtool/blob/master/docker/owlbot/java/src/requirements.txt) to the latest version now that we're using Python 3.9. It'll help us resolve these open PRs: https://github.com/googleapis/synthtool/pull/1480, https://github.com/googleapis/synthtool/pull/1731. 

And for `requirements.in`, mostly a clean up but afaik, pinning is only required to freeze the version/prevent a dep from being upgraded to the latest. This may not be needed anymore with the Python upgrade so we should be fine with unpinning the versions there. wdyt?

lqiu96:
> > @mpeddada1 What did you mean by update? Should I just unpin the dependencies or pin the current latest versions or something else?
> 
> I think we can now update the dependencies in the associated [requirement.txt](https://github.com/googleapis/synthtool/blob/master/docker/owlbot/java/src/requirements.txt) to the latest version now that we're using Python 3.9. It'll help us resolve these open PRs: #1480, #1731.
> 
> And for `requirements.in`, mostly a clean up but afaik, pinning is only required to freeze the version/prevent a dep from being upgraded to the latest. This may not be needed anymore with the Python upgrade so we should be fine with unpinning the versions there. wdyt?

Sounds good. I'll unpin the versions

mpeddada1 (approved):

lqiu96:
Adding `Do Not Merge` to this PR. Plan is to merge after the next release of the libraries-bom.

blakeli0:
> And for requirements.in, mostly a clean up but afaik, pinning is only required to freeze the version/prevent a dep from being upgraded to the latest

@mpeddada1 Did you validate this approach with other languages lead or rennie, based on his "Pip install vulnerability" doc, I think we may still need it.

docker/owlbot/java/container_test.yaml
```diff
@@ -23,4 +23,4 @@ commandTests:
   expectedError: ["google-java-format: Version 1.7"]
 - name: "python"
   command: ["python", "--version"]
-  expectedOutput: ["Python 3.6.1"]
+  expectedOutput: ["Python 3.9.13"]
```
suztomo, line 26: @lqiu96 Would you add an assertion that this python module `synthtool.languages.java` is available? I think it would look like this:

```
python3 -c 'import synthtool.languages.java; print("import successs")'
```

An example failure would look like this:

```
suztomo@suztomo:~/java-storage$ docker run --user "$(id -u):$(id -g)" --entrypoint /bin/bash --rm -it  a8e0ab6765a4
I have no name!@752c3d65c2dd:/workspace$ python3 -c 'import synthtool.languages.java; print("import successs")'
Traceback (most recent call last):
  File "<string>", line 1, in <module>
ModuleNotFoundError: No module named 'synthtool.languages'
I have no name!@752c3d65c2dd:/workspace$ 
```

It gives more confidence that a change for OwlBot Postprocessor works fine in owlbot.py ([example location](https://github.com/googleapis/google-cloud-java/blob/main/java-accessapproval/owlbot.py#L16))

lqiu96: Sure, I can add that validation in

lqiu96: Added the new validation in as `java synthtool validation` which runs the commands above.

FYI, couldn't find the docs for `schemaVersion` 1.0.0 and the docs on https://github.com/GoogleContainerTools/container-structure-test were showing docs for `2.0.0`, so I used that for local testing

mpeddada1:
> > And for requirements.in, mostly a clean up but afaik, pinning is only required to freeze the version/prevent a dep from being upgraded to the latest
> 
> @mpeddada1 Did you validate this approach with other languages lead or rennie, based on his "Pip install vulnerability" doc, I think we may still need it.

@blakeli0 sorry I missed your comment earlier! Yup, according to the document `pip install --require-hashes -r /synthtool/requirements.txt` line ensures that we mitigate the vulnerability. The pinning of versions was a java-specific issue and we needed it because we were still using an older version of Python that wasn't always compatible with the latest versions of the dependencies. Languages like Python don't pin the versions (for example: https://github.com/googleapis/synthtool/blob/master/synthtool/gcp/templates/python_library/.kokoro/requirements.in).

docker/owlbot/java/container_test.yaml
suztomo: Thank you.

suztomo (approved):

lqiu96:
Removing the `Do Not Merge` label... Planning on releasing python v3.9.

chingor13 (approved):

PR #1753, opened 2023-01-20
chore(python): exclude grpcio==1.52.0rc1 in tests

Mariatta:
grpcio 1.52.0rc1 is causing system tests to hang indefinitely, with the following message:
```
E0120 09:47:39.204036000 4390417920 thread_pool.cc:254]                Waiting for thread pool to idle before forking
```

Exclude it from client libraries.

See https://github.com/grpc/grpc/issues/32163


parthea (approved):
Thank you @Mariatta !

PR #1757, opened 2023-01-28
chore(java): omit `-Penable-integration-tests` parameter from native image test run command

mpeddada1:
The opting in of tests for native image testing is explicitly done in the `native` profile's maven-surefire-plugin. See https://github.com/googleapis/java-shared-config/blob/44a77b0faf5b4f3f9208117ccab0a3717d059efd/pom.xml#L806-L817 for reference. Therefore, the `-Penable-integration-tests` profile is redundant.

### DIFF
--- a/synthtool/gcp/templates/java_library/.kokoro/build.sh
+++ b/synthtool/gcp/templates/java_library/.kokoro/build.sh
@@ -71,12 +71,12 @@ integration)
     ;;
 graalvm)
     # Run Unit and Integration Tests with Native Image
-    mvn -B ${INTEGRATION_TEST_ARGS} -ntp -Pnative -Penable-integration-tests test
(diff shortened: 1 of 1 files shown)


BenWhitehead (approved):
An additional point of clarity:
Normally (for non-native builds), our config for surefire (and the `test` phase) will exclude integration tests from running, deferring those integration tests to failsafe and the `integration-test` phase. In the case of the `native` profile, we actually exclude unit tests from being run, and shift the integration tests patter from `integration-test` instead to `test` which `native` then uses.

Despite our change here saying we're not running integration tests, we in fact are and have been running them in the `test` phase for some time.

mpeddada1:
Thanks for the additional context @BenWhitehead!

PR #1763, opened 2023-02-08
feat: python accept passed files to exclude

leahecole:
Add capability for Python to accept a list of files to exclude, as suggested by @parthea


leahecole:
force push was to amend the commit author - my bad!

parthea (approved):

PR #1770, opened 2023-02-22
chore: generate non-cloud client library

JoeWang1127:
Fix https://github.com/googleapis/google-cloud-java/issues/9125


docker/owlbot/java/src/fix-poms.py
```diff
@@ -274,6 +274,12 @@ def update_bom_pom(filename: str, modules: List[module.Module]):
 
     tree.write(filename, pretty_print=True, xml_declaration=True, encoding="utf-8")
 
+
+def _proto_group_id(group_id: str) -> str:
```
suztomo, line 278:
```suggestion
def _proto_group_id(main_artifact_group_id: str) -> str:
```

suztomo, line 278: Can you add source code comment why we had to split the proto group IDs?

docker/owlbot/java/tests/new-client-maps/golden/grpc-google-maps-foo-v1/pom.xml
```diff
@@ -0,0 +1,69 @@
+<project xmlns="http://maven.apache.org/POM/4.0.0"
+  xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
+  xsi:schemaLocation="http://maven.apache.org/POM/4.0.0 http://maven.apache.org/xsd/maven-4.0.0.xsd">
+  <modelVersion>4.0.0</modelVersion>
+  <groupId>com.google.maps.api.grpc</groupId>
```
suztomo, line 5: Nice test.

suztomo (approved):
Small suggestion for comment and variable name. Approving this pull request regardless whether you incorporate it or not.

docker/owlbot/java/src/fix-poms.py
JoeWang1127: done

JoeWang1127: done

PR #1771, opened 2023-02-27
feat(node): add default cred service account for tests

alvarowolfx:
For this new feature on the BigQuery Storage API client, I was trying to write tests that hit the actual service running, as the service uses aBiDi gRPC communication, has a complex interaction pattern and is just more feasible to just test against the real thing ( we do that on the BigQuery Go SDK ). 

The problem is that today, test scripts on CI for Node doesn't have the service account credentials there, but they do exist for samples and system tests. This PR adds the service account credentials to the test scripts.

PR that is blocked by this: https://github.com/googleapis/nodejs-bigquery-storage/pull/328

I wasn't able to test this on Windows, not entirely sure if what I wrote is gonna work correct fro that environment.

### DIFF
--- a/synthtool/gcp/templates/node_library/.kokoro/continuous/node12/test.cfg
+++ b/synthtool/gcp/templates/node_library/.kokoro/continuous/node12/test.cfg
@@ -1,0 +1,4 @@
+env_vars: {
+  key: "SECRET_MANAGER_KEYS"
+  value: "long-door-651-kokoro-system-test-service-account"
+}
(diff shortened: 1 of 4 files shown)


chingor13:
IMO, unit tests should not be need credentials, but deferring to @sofisl to decide for nodejs.

alvarowolfx:
agree with you @chingor13, unit tests shouldn't touch servers. But how to run/handle integration tests ? 

I'm new on development in node SDKs here, I came from the go cloud SDKs and it's pretty normal for us to run integrations tests like that, specially for complex interactions where mocking a lot of parts of the system just takes too much work and ends up not being that accurate.

sofisl:
Hi @alvarowolfx, I think it's fine if you set up your unit tests to run with credentials, but I'd only do it for your library. Just add the files you want to exclude [here](https://github.com/googleapis/nodejs-bigquery-storage/blob/a516d9ab160bb553bb8f215cacd6549507608e3d/owlbot.py#L17), and make the changes right on the library itself.

alvarowolfx:
I checked some other libs, like BigTable, and saw that for integration tests `system-test` is used. Those have access to the credentials so this change is not needed. Thanks for the tips and sorry for the noise.

PR #1777, opened 2023-03-16
chore: treat java-iam as special structure module

suztomo:


suztomo:
Java-iam was recently moved to google-cloud-java.

PR #1787, opened 2023-04-03
chore: make exception for non-cloud artifacts

JoeWang1127:
Fix https://github.com/googleapis/google-cloud-java/issues/9304

### DIFF
--- a/docker/owlbot/java/src/fix-poms.py
+++ b/docker/owlbot/java/src/fix-poms.py
@@ -283,9 +283,10 @@ def update_bom_pom(filename: str, modules: List[module.Module]):
 # https://github.com/googleapis/google-cloud-java/issues/9304
 def _proto_group_id(main_artifact_group_id: str) -> str:
     prefix = "com.google"
-    if main_artifact_group_id != "com.google.cloud" and \
-       main_artifact_group_id != "com.google.area120" and \
-       main_artifact_group_id != "com.google.analytics":
+    list_of_group_id = ["com.google.cloud",
+                        "com.google.area120",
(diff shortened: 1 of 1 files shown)


docker/owlbot/java/src/fix-poms.py
```diff
@@ -279,10 +279,14 @@ def update_bom_pom(filename: str, modules: List[module.Module]):
...
+# https://github.com/googleapis/google-cloud-java/issues/9304
 def _proto_group_id(main_artifact_group_id: str) -> str:
-    prefix = "com.google" \
-        if main_artifact_group_id == "com.google.cloud" \
-        else main_artifact_group_id
+    prefix = "com.google"
+    if main_artifact_group_id != "com.google.cloud" and \
+       main_artifact_group_id != "com.google.area120" and \
+       main_artifact_group_id != "com.google.analytics":
+        prefix = main_artifact_group_id
```
JoeWang1127, line 289: Python 3.9 doesn't have a `switch` statement.

suztomo: Can you use "in" operator with a list of the 3 elements?

suztomo: Also add source code comment for the reason why this logic is needed.

suztomo (changes requested):
@JoeWang1127 Update the title with "chore:" and ensure you merge with "chore:" to main. This line may appear in the pull request in downstream repositories and without "chore:" it would appear in their release note of irrelevant repositories.

JoeWang1127:
@suztomo changed the PR title.

PR #1790, opened 2023-04-11
build: update README links

sofisl:


chingor13 (changes requested):
I really suggest having a test for your post-processor image. It's not easy to tell if this is going to work or not.

Take a look at how we do the golden tests for the java post-processor image.

sofisl:
That's fair. I've created a tracking issue so as to not pin this work on that:

https://github.com/googleapis/google-cloud-node/issues/4165

chingor13 (approved):

PR #1795, opened 2023-05-03
test: add tests for readme generation and address some bugs

bcoe:
* add some tests for README generation step.
* Files in samples/test/foo were included in README samples, which is confusing because we link to tests as if they're a sample..
* full_directory_path included the googleapis/google-cloud-node, introduced directory_path, which includes only the path within google-cloud-node.


synthtool/languages/node_mono_repo.py
```diff
@@ -57,11 +57,15 @@ def read_metadata(relative_dir: str):
         )
 
         repo = git.parse_repo_url(repo_url)
-
+        data["directory_path"] = (
+            data["repository"]
+            if isinstance(data["repository"], str)
+            else f'{data["repository"]["directory"]}'
```
sofisl, line 63: The existence of the `repository.directory` field was a later addition, so it might throw an error if it doesn't exist. Perhaps better to add another fallback.

sofisl (approved):

PR #1797, opened 2023-05-10
chore(python): remove release post processing for python monorepo

parthea:


synthtool/languages/python_mono_repo.py
```diff
@@ -222,55 +181,8 @@ def owlbot_main(package_dir: str) -> None:
...
-                "extra-files": output,
-            }
-
-    with open(release_please_config, "w") as f:
-        json.dump(config_json, f, indent=4)
-
-
 if __name__ == "__main__":
     owlbot_dirs = walk_through_owlbot_dirs(Path.cwd())
-    configure_release_please(owlbot_dirs)
```
vchudnov-g, line 273: As per our discussion today: for now, rather than deleting the code in lines 225-270, I would just comment out this invocation in 273 until we have a replacement mechanism in place. This way, if we need this functionality after the monorepo-migration but before the new mechanism is in place, we can call this code again.

Actually, this suggests another idea: maybe we can have this call [and the `create_gapic_version` call] only occur if the PR title does not begin with "chore(migration)", or some other sentinel. That way this tool just continues to work as currently in all other cases, which would be ideal until the replacement mechanism is fully in place. WDYT?

synthtool/languages/python_mono_repo.py
```diff
@@ -18,44 +18,6 @@
...
-# See the License for the specific language governing permissions and
-# limitations under the License.
-#
-
-__version__ = "0.0.0"  # {x-release-please-version}
-"""
-
-    if not gapic_version_path.exists():
-        with open(gapic_version_path, "w") as f:
-            f.write(gapic_version_text)
```
vchudnov-g, line 58: Consider keeping all this code and only commenting out the invocation. See my other comment below, particularly the idea about another way to selectively skip calling this code

vchudnov-g: See overall review comment.

synthtool/languages/python_mono_repo.py
vchudnov-g: See overall review comment.

vchudnov-g (approved):
I really like the idea of calling the code when we're not dealing with migration PRs, so the rest of the process keeps working.

As per our offline conversation, though, you think the replacement tool will be available shortly and don't want to complicate this PR. I'm grudgingly fine with that as long as (a) we clearly document the extra steps we need to take to release new libraries (or whether we should have a freeze on that!) until the new tool is ready, and (b) the tool lands soon so we can bring our toil and tech debt back down.

Consider merging this PR as close as makes sense to having the tool ready, if that helps minimize disruption to our process.

parthea:
The tool will be available today. Docs will also be updated today. I'll keep this PR open until I have the CL and tool ready.

parthea:
Created cl/535319618 to update the documentation for new client library generation using the new tool.

Marked PR https://github.com/googleapis/google-cloud-python/pull/11181 as ready for review which contains the code for the new tool.

vchudnov-g (approved):

PR #1799, opened 2023-05-15
build: migrate release scripts

chingor13:
We are migrating the Java release scripts into shared versions in piper. This change needs to merge with the corresponding internal changelist for the Kokoro job config.


suztomo:
Would you create a spreadsheet or automation to capture the Kokoro job status and this multiscmname update? I believe we will have some gaps.

suztomo (approved):

PR #1810, opened 2023-06-06
fix: update copyright year in Java license header replacement template

emmileaf:
Following up to Java generator's copyright year fix and discussion in https://github.com/googleapis/sdk-platform-java/pull/1720#issuecomment-1568919167 with post-processor fix.

Todos:
- [x] Fix copyright year
- [x] Update golden files ([instructions](https://github.com/googleapis/synthtool/tree/master/docker/owlbot/java#rebuilding-golden-test-fixtures)) for Java OwlBot Image Tests
- [x] Update [golden files](https://github.com/googleapis/synthtool/tree/master/tests/testdata) for tests in `test_language_java.py`
- [x] Check locally built docker image against java-storage (split repo) - no changes to existing files
- [x] Check locally built docker image against google-cloud-java (monorepo) - no changes to existing files


dandhlee:
FYI this is optional and not required to update every year - copyright should reflect when these changes were first introduced, and it's better to have coverage to older years than risk not having potential coverage for previous years.

Generated files _can_ be updated, while other files should not have it license updated.

suztomo (approved):

emmileaf:
Ah that makes sense, thanks for the heads up @dandhlee! 

For a bit of context - the intention of this change was to fix the header years in newly generated files where this logic is applied to (e.g. in https://github.com/googleapis/google-cloud-java/pull/9480), so that their headers have 2023 rather than 2020). 

That said, is the concern here that the change might unintentionally modify headers on existing files as part of the generation process? Maybe a better fix might be to parse out the year from “BAD_LICENSE” and use that instead of a hard-coded value though - I can take a look at that approach as an alternative here.

dandhlee:
For autogenerated files either is fine - they can be 2020 or 2023, doesn't matter (as long as we're consistent about it). Existing files can also get copyright updates if they're autogenerated files. 

Just want to let you know that this is an optional change and it is not strictly necessary that these files have to say they're in the current year.

parthea (approved):

PR #1811, opened 2023-06-13
chore: [autoapprove] Update `black` and `isort` to latest versions

holtskinner:
- Add Google Profile to `isort` `format` session
- Add `black[jupyter]` because several samples directories include jupyter notebooks


synthtool/gcp/templates/python_library/noxfile.py.j2
```diff
@@ -179,7 +182,10 @@ def format(session):
     # See https://pycqa.github.io/isort/docs/configuration/options.html#force-sort-within-sections
     session.run(
         "isort",
+        "--profile=google",
         "--fss",
+        "--known-local-folder=main",
+        "--treat-all-comment-as-code",
```
parthea, line 188: This will cause a lot of changes in the generated libraries. Please create a doc to discuss the motivation for the change.

synthtool/gcp/templates/python_library/noxfile.py.j2
```diff
@@ -17,21 +17,23 @@
...
 import re
 import shutil
+from typing import Dict, List
 import warnings
 
 import nox
 
-BLACK_VERSION = "black==22.3.0"
-ISORT_VERSION = "isort==5.10.1"
+BLACK_VERSION = "black[jupyter]==23.3.0"
```
parthea, line 30: Please also update `black` here
https://github.com/googleapis/synthtool/blob/master/synthtool/gcp/templates/python_library/.pre-commit-config.yaml#L24

Can you clarify why the `jupyter` extra is needed?

synthtool/gcp/templates/python_library/noxfile.py.j2
parthea: The changes to this noxfile should also be in `synthtool/gcp/templates/python_mono_repo_library/noxfile.py.j2` for consistency between libraries in the split repositories vs monorepo.

parthea (commented):
Added observations

synthtool/gcp/templates/python_library/noxfile.py.j2
holtskinner: The jupyter extra is needed for repositories that contain jupyter notebooks like https://github.com/GoogleCloudPlatform/vertex-ai-samples/ and https://github.com/GoogleCloudPlatform/generative-ai

synthtool/gcp/templates/python_library/noxfile.py.j2
holtskinner: I'm going to revert this change, after experimentation, I realized it wouldn't improve the results.

synthtool/gcp/templates/python_library/noxfile.py.j2
holtskinner: Actually, on second thought. Should I add the `jupyter` extra to the python samples noxfile instead?

holtskinner:
Cleaned up commit history.

parthea (approved):

PR #1818, opened 2023-06-27
build(deps): Address dependabot security vulnerabilities

holtskinner:
- Add version restrictions into `requirements.in` in addition to `requirements.txt`
- Replaces 
  - #1801
  - #1802
  - #1807


parthea:
Let’s keep the other PRs instead of grouping them together

PR #1847, opened 2023-08-02
build: update Node.js version and base Debian version

alexander-fenster:
Node.js v15 is EOL, and the newest version of `eslint` requires something newer since it uses new features (namely, `require("node:fs")` which does not work in Node 15.

While I'm here, I'm also updating the Debian base image to `bookworm` (`buster` is EOL), and fixing some typos in the comments.

Last but not least, let's not do `curl | bash` in Dockerfiles.

### DIFF
--- a/docker/owlbot/nodejs/Dockerfile
+++ b/docker/owlbot/nodejs/Dockerfile
@@ -12,17 +12,19 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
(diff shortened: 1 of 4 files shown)


alexander-fenster:
It seems I'm unable to see what's going on with the CI.

suztomo (approved):

parthea (approved):

bcoe (approved):

PR #1850, opened 2023-08-17
fix: update release-please post-processing for nodejs apiary

sofisl:


synthtool/languages/node.py
```diff
@@ -278,6 +291,18 @@ def _noop(library: Path) -> None:
     pass
 
 
+def walk_through_apiary(dir, glob_to_search_for):
```
bcoe, line 294: A couple comments might be worthwhile:

`# walk through apiary to determine the current list of APIs that have been generated`

synthtool/languages/node.py
```diff
@@ -270,6 +270,19 @@ def postprocess_gapic_library_hermetic(hide_output=False):
     logger.debug("Post-processing completed")
 
 
+def write_release_please_config(dirs: list):
```
bcoe, line 273: I think a comment would be worthwhile here explaining that we are adding default empty `{}` entries for every API that we find during the running of `walk_through_apiary`.

tests/test_node.py
```diff
@@ -240,6 +242,35 @@ def test_owlbot_main_with_staging_index_from_staging(hermetic_mock, nodejs_dlp):
     assert staging_text == text
 
 
+def test_write_release_please_config():
```
bcoe, line 245: great tests!

bcoe (approved):
LGTM.

Could you update the issue body to explain what this fixes? I believe the issue is that new APIs haven't had release please config?

PR #1876, opened 2023-10-03
chore: track customization in templates

### LINKED ISSUE #1798: Owlbot should keep track of custom modifications to templates
We have some examples of libraries that make modifications to otherwise standard templates. For example, https://github.com/googleapis/java-bigtable/blob/main/.kokoro/presubmit/integration.cfg#L9-L12 adds custom IT args to this standard template: https://github.com/googleapis/synthtool/blob/master/synthtool/gcp/templates/java_library/.kokoro/presubmit/integration.cfg. As such, we need to exclude it from owlbot and any new iterations of these templates need to be handled manually (see https://github.com/googleapis/java-bigtable/pull/1255 and fix https://github.com/googleapis/java-bigtable/pull/1258).

It would be great if owlbot could preserve known additions to templates so we can still receive updates without having to do manual work on them in a repeated way.

JoeWang1127:
## Append env_vars in *all* kokoro build configuration files (*.cfg in .kokoro/) which contain an `env_vars` (key: JOB_TYPE, value: integration)

When you want to append the environment variable too *all* templated .cfg which has an env_vars (key: JOB_TYPE, value: integration), follow the steps below: 

1. Create a partial file in the repository root directory. 
   Note that the file extension should be `.yaml`.
2. Write customization entries in the partial file.
3. Create a pull request containing changes in step 1 - 2
4. Confirm the customizations appear in the configuration file in the same pull request after the OwlBot post processor runs.

## Modify default env_vars in Kokoro build templates

You don’t need to modify existing env_vars defined in the templates because if there are multiple env_vars with the same key, the last value will be used.

Please refer to [Maps](https://protobuf.dev/programming-guides/proto3/#maps) in proto 3 language guide.

### Append env_vars in configuration files with an env_vars (key: JOB_TYPE, value: integration)
1. Add a partial file `.integration-partials.yaml` to the repository root.
2. In `.integration-partials.yaml`, add the following section:
```
integration: |
  env_vars: {
    key: "INTEGRATION_TEST_ARGS"
    value: "-P bigtable-emulator-it"
  }
  
  env_vars: {
    key: "GCLOUD_PROJECT"
    value: "gcloud-devel"
  }

```
3. Create a pull request containing the changes in step 1 - 2.
### After the change
The following section will be appended to 
- `.kokoro/nightly/integration.cfg`
- `.kokoro/nightly/java11-integration.cfg`
- `.kokoro/presubmit/integration.cfg`
```
env_vars: {
  key: "INTEGRATION_TEST_ARGS"
  value: "-P bigtable-emulator-it"
}

env_vars: {
  key: "GCLOUD_PROJECT"
  value: "gcloud-devel"
}

```
Because these files contain env_vars (key: JOB_TYPE, value: integration)

For env_var entry with key GCLOUD_PROJECT, the default value is overwritten by customized value (gcloud-devel) that appears later in the file.

Fix #1798.


JoeWang1127:
Kokoro CI error:
```
E               subprocess.CalledProcessError: Command '['docker', 'build', '-t', 'owlbot-php-test', '-f', 'docker/owlbot/php/Dockerfile', '.']' returned non-zero exit status 1.

/[usr/local/lib/python3.8/subprocess.py:516](https://cs.corp.google.com/piper///depot/google3/usr/local/lib/python3.8/subprocess.py?l=516): CalledProcessError
```

JoeWang1127:
Changed test directory and only lint failed:
```
would reformat /tmpfs/src/github/synthtool/synthtool/gcp/common.py
would reformat /tmpfs/src/github/synthtool/synthtool/languages/java.py
would reformat /tmpfs/src/github/synthtool/tests/test_language_java.py
```

synthtool/gcp/templates/java_library/.kokoro/nightly/integration.cfg
```diff
@@ -1,37 +1,31 @@
 # Format: //devtools/kokoro/config/proto/build.proto
 
-# Configure the docker image for kokoro-trampoline.
+{%- if 'defaults' in metadata and 'kokoro_nightly_env_vars' in metadata['defaults'] -%}
```
suztomo, line 3: It looks as if the logic only works for kokoro nightly builds environment variables. Does this meet Kristen's issue?

suztomo: I'll send a separate message.

synthtool/gcp/partials.py
```diff
@@ -34,13 +37,15 @@ def load_partials(files: List[str] = _DEFAULT_PARTIAL_FILES) -> Dict:
     deprecation_warning: a warning to indicate that the library has been
         deprecated and a pointer to an alternate option
     """
+    if files is None:
+        files = _DEFAULT_PARTIAL_FILES
+    else:
+        files.extend(_DEFAULT_PARTIAL_FILES)
+    res = {}
```
SurferJeffAtGoogle, line 44: res == result?  Mind renaming it result?

SurferJeffAtGoogle (dismissed):

synthtool/gcp/partials.py
JoeWang1127: done

synthtool/gcp/partials.py
```diff
@@ -34,13 +37,15 @@ def load_partials(files: List[str] = _DEFAULT_PARTIAL_FILES) -> Dict:
     deprecation_warning: a warning to indicate that the library has been
         deprecated and a pointer to an alternate option
     """
+    if files is None:
+        files = _DEFAULT_PARTIAL_FILES
+    else:
+        files.extend(_DEFAULT_PARTIAL_FILES)
```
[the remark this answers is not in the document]
JoeWang1127: done

synthtool/gcp/common.py
```diff
@@ -49,10 +49,33 @@ def __init__(self, template_path: Optional[Path] = None):
         self._templates = templates.Templates(self._template_root)
         self.excludes = []  # type: List[str]
 
-    def _generic_library(self, directory: str, relative_dir=None, **kwargs) -> Path:
```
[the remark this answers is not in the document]
JoeWang1127: Thanks for the suggestion, done.

synthtool/gcp/partials.py
```diff
@@ -21,9 +21,9 @@
 _DEFAULT_PARTIAL_FILES = [".readme-partials.yml", ".readme-partials.yaml"]
```
suztomo, line 21: What's the reason why you didn't choose to add `.integration-partials.yaml` to this `_DEFAULT_PARTIAL_FILES`? 

My first impression is that the change would be simpler.

tests/fixtures/java_templates/partials_test/.integration-partials.yaml
```diff
@@ -0,0 +1,10 @@
+integration: |
```
suztomo, line 1: I want the semantic of this value coming at the end of the generated cfg file. Can you add "append" or something at the key name?

synthtool/gcp/common.py
```diff
@@ -321,11 +321,32 @@ def py_library(self, **kwargs) -> Path:
...
         return ret
 
-    def java_library(self, **kwargs) -> Path:
+    def java_library(self, partial_files: List[str] = [], **kwargs) -> Path:
         # kwargs["metadata"] is required to load values from .repo-metadata.json
         if "metadata" not in kwargs:
             kwargs["metadata"] = {}
-        return self._generic_library("java_library", **kwargs)
+        # load common repo meta information (metadata that's not language specific).
+        if "metadata" in kwargs:
```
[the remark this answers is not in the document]
JoeWang1127: I followed @SurferJeffAtGoogle suggestion [here](https://github.com/googleapis/synthtool/pull/1876#discussion_r1346451926).

synthtool/gcp/partials.py
JoeWang1127: I think customer may want to explicitly use partials in `java.common_templates`.

Even if I added `.integration-partials.yaml` to `_DEFAULT_PARTIAL_FILES`, the logic still needs change because previously only one partial file will be loaded.

tests/fixtures/java_templates/partials_test/.integration-partials.yaml
suztomo: I found rennie's comment: https://github.com/googleapis/synthtool/pull/1876#discussion_r1346451926

JoeWang1127: Changed key to `integration_append`.

synthtool/gcp/partials.py
suztomo: > the logic still needs change because previously only one partial file will be loaded.

Yes, I think that's valid.

```
    for file in files + _DEFAULT_PARTIAL_FILES:
        partials_file = cwd_path / file
        if os.path.exists(partials_file):
            with open(partials_file) as f:
                result.update(yaml.load(f, Loader=yaml.SafeLoader))
```

> I think customer may want to explicitly use partials in java.common_templates.

I don't feel so. But it's not strong. Think about that (possibility of not needing `partial_files` arguments to  `java_library` and `_load_generic_metadata`) again. If you still feel it's better, then I'm good.

suztomo (approved):
Think about the possibility of not needing partial_files arguments to java_library and _load_generic_metadata. If you still feel it's better to let users to explicitly specify the partial files, then I'm good. Approving either case.

synthtool/gcp/common.py
```diff
@@ -321,11 +321,32 @@ def py_library(self, **kwargs) -> Path:
...
+        t = templates.TemplateGroup(self._template_root / "java_library", self.excludes)
+
+        if "repository" in kwargs["metadata"] and "repo" in kwargs["metadata"]:
+            kwargs["metadata"]["repo"]["default_branch"] = _get_default_branch_name(
+                kwargs["metadata"]["repository"]
+            )
+
+        result = t.render(**kwargs)
+        _tracked_paths.add(result)
+
```
chingor13, line 348: Can this logic go into the `java.common_templates()` call instead?

Context: when this code was originally written, the inheritance was backwards (this common template code had methods for each language and they were explicitly used in synth.py files in each repository). The intention was to migrate them all to use `<lang>.common_templates()` or something like that and then remove these language-specific methods on this class.

chingor13 (changes requested):
Generally, looks ok to me

synthtool/gcp/partials.py
JoeWang1127: I decided not to change the parameter of `java_library` and `_load_generic_metadata` to reduce the complexity of this PR.

synthtool/gcp/common.py
JoeWang1127: I decided not to change the interface of `java.common_templates` and `java_library` as this refactoring should go to a dedicated PR.

chingor13 (approved):

PR #1881, opened 2023-10-17
chore(python_mono_repo): add support for client specific post processing

parthea:
This PR adds support for client specific post processing in the python owlbot monorepo post processor image.


synthtool/languages/python_mono_repo.py
```diff
@@ -98,6 +99,54 @@ def update_url_in_setup_py(package_dir: str):
...
+    """Applies client specific post processing which exists in
+    the Path `client_specific_post_processing_dir`. The client specific post processing
+    YAML is in the following format:
+    ```
+        description: Verbose description about the need for the workaround.
+        url: Url of the issue in gapic-generator-python.
+        replacements:
+          - replacement:
+            paths: [<List of files where the replacement should occur>]
+            before: ""The string to search for in the specified paths""
```
vchudnov-g, line 112:
```suggestion
            before: "The string to search for in the specified paths"
```

synthtool/languages/python_mono_repo.py
```diff
@@ -98,6 +99,54 @@ def update_url_in_setup_py(package_dir: str):
         f.writelines(new_setup_py)
 
 
+def apply_client_specific_post_processing(client_specific_post_processing_dir: Path):
+    """Applies client specific post processing which exists in
+    the Path `client_specific_post_processing_dir`. The client specific post processing
+    YAML is in the following format:
+    ```
+        description: Verbose description about the need for the workaround.
+        url: Url of the issue in gapic-generator-python.
```
vchudnov-g, line 108:
```suggestion
        url: URL of the issue in gapic-generator-python tracking eventual removal of the workaround
```

synthtool/languages/python_mono_repo.py
```diff
@@ -98,6 +99,54 @@ def update_url_in_setup_py(package_dir: str):
         f.writelines(new_setup_py)
 
 
+def apply_client_specific_post_processing(client_specific_post_processing_dir: Path):
+    """Applies client specific post processing which exists in
```
vchudnov-g, line 103:
```suggestion
    """Applies client-specific post processing which exists in
```

synthtool/languages/python_mono_repo.py
```diff
@@ -98,6 +99,54 @@ def update_url_in_setup_py(package_dir: str):
...
+    the Path `client_specific_post_processing_dir`. The client specific post processing
```
vchudnov-g, line 104:
```suggestion
    the Path `client_specific_post_processing_dir`. The client-specific post processing
```

synthtool/languages/python_mono_repo.py
```diff
@@ -98,6 +99,54 @@ def update_url_in_setup_py(package_dir: str):
...
+          - replacement:
+            paths: [<List of files where the replacement should occur>]
+            before: ""The string to search for in the specified paths""
+            after:  "The string to replace in the the specified paths",
+            count: <integer indicating number of replacements that should have occurred across all files after the script is run>
+    ```
+
+    Args:
+        client_specific_post_processing_dir (str): path to the directory which contains YAML files which will
+            be used to apply client specific post processing. For example, 'scripts/client-post-processing'
```
vchudnov-g, line 119:
```suggestion
            be used to apply client-specific post processing, e.g. 'scripts/client-post-processing'
```

synthtool/languages/python_mono_repo.py
```diff
@@ -98,6 +99,54 @@ def update_url_in_setup_py(package_dir: str):
...
+def apply_client_specific_post_processing(client_specific_post_processing_dir: Path):
+    """Applies client specific post processing which exists in
+    the Path `client_specific_post_processing_dir`. The client specific post processing
+    YAML is in the following format:
+    ```
+        description: Verbose description about the need for the workaround.
+        url: Url of the issue in gapic-generator-python.
+        replacements:
+          - replacement:
+            paths: [<List of files where the replacement should occur>]
```
vchudnov-g, line 111: Specify the root dir for this path. I assume the repo root?

synthtool/languages/python_mono_repo.py
```diff
@@ -98,6 +99,54 @@ def update_url_in_setup_py(package_dir: str):
...
+                for replacement in replacements:
+                    # For each file that needs the workaround applied
+                    for client_library_path in replacement["paths"]:
+                        assert (
+                            s.replace(
+                                client_library_path,
+                                replacement["before"],
+                                replacement["after"],
+                            )
+                            == replacement["count"]
```
vchudnov-g, line 137: Doesn't this violate idempotency? We'll make the replacements in all the package dirs the first time. Then when this script is triggered again, we'll only make changes in the dirs that changed; the other ones will get no changes, so there will be no replacements, so the assertion will fail.

In the design doc we said that `replacement["count"]` will be used to check the difference between the original and changed files, so it seems like we need to somehow cache the original file before it's ever processed by this script, and then we can compare it to the what it becomes after this script has run any number of times.

Maybe you mean to only selectively trigger this script? But then perhaps it would make sense to implement the symlink idea I mentioned in the design doc. (My thinking: if the workaround file applies to packages A and B and only B has new changes, the assertion will fail on A.)

Happy to discuss more, especially if my interpretation/mental model is incorrect.

synthtool/languages/python_mono_repo.py
```diff
@@ -98,6 +99,54 @@ def update_url_in_setup_py(package_dir: str):
...
+            be used to apply client specific post processing. For example, 'scripts/client-post-processing'
+    """
+    if Path(client_specific_post_processing_dir).exists():
+        for post_processing_path in Path(client_specific_post_processing_dir).iterdir():
+            with open(post_processing_path, "r") as post_processing_path_file:
+                post_processing_json = yaml.safe_load(post_processing_path_file)
+                replacements = post_processing_json["replacements"]
+
+                # For each workaround related to the specified issue
+                for replacement in replacements:
```
vchudnov-g, line 128: Suggest s/`replacements`/`all_replacements`/ so we don't have variabless differing by  a letter

synthtool/languages/python_mono_repo.py
```diff
@@ -98,6 +99,54 @@ def update_url_in_setup_py(package_dir: str):
...
+                            s.replace(
+                                client_library_path,
+                                replacement["before"],
+                                replacement["after"],
+                            )
+                            == replacement["count"]
+                        )
+                        # Ensure that subsequent calls won't trigger additional replacements
+                        assert (
+                            s.replace(
```
vchudnov-g, line 141: nit nit: IMO it would be clearer to s/`s`/`synthtool`/ via the import above. I know it's been like this before this PR.

synthtool/languages/python_mono_repo.py
parthea: Fixed in c525b70

synthtool/languages/python_mono_repo.py
parthea: You're right. I meant to trigger the replacement selectively. I've updated the design doc to include symlinks under package specific folders, and implemented the idea in this PR. Please take another look at the design doc and changes in [c525b70](https://github.com/googleapis/synthtool/pull/1881/commits/c525b700e68b22e19fda4ae5544638d9dab2fe82).

synthtool/languages/python_mono_repo.py
parthea: Fixed in [cc273d8](https://github.com/googleapis/synthtool/pull/1881/commits/cc273d86eb61a816d0cd72d7ca12cdb9d04a9654)

synthtool/languages/python_mono_repo.py
parthea: Fixed in a2afa61

synthtool/languages/python_mono_repo.py
```diff
@@ -98,6 +99,60 @@ def update_url_in_setup_py(package_dir: str):
...
+
+    Args:
+        client_specific_post_processing_dir (str): Path to the directory which contains YAML files which will
+            be used to apply client-specific post processing, e.g. 'packages/<package_name>/scripts/client-post-processing'
+            relative to the monorepo root directory.
+        package_name (str): The name of the package where client specific post processing will be applied.
+    """
+
+    if Path(client_specific_post_processing_path).exists():
+        for post_processing_path in Path(
```
vchudnov-g, line 126: nit: can we have the condition in just one line for readability?

synthtool/languages/python_mono_repo.py
```diff
@@ -98,6 +99,60 @@ def update_url_in_setup_py(package_dir: str):
...
+                    # For each file that needs the workaround applied
+                    for client_library_path in replacement["paths"]:
+                        if package_name in client_library_path:
+                            assert (
+                                synthtool.replace(
+                                    client_library_path,
+                                    replacement["before"],
+                                    replacement["after"],
+                                )
+                                == replacement["count"]
```
vchudnov-g, line 143: This still violates idempotency, because `replacement["count"]` is meant to be the sum of all replacements across all the files in `replacement["paths"]`, not just under `package_name`. If we keep those semantics, the assert should not be here, and instead should refer to a (cached?) version of the raw generated files vs the post-processed files.

[Potentially we should change the semantics of `replacement["count"]` to be per `client_library_path`, but looking at the source of `synthtool.replace`, the first argument could be a glob and so the number of replacements could vary by package.]

synthtool/languages/python_mono_repo.py
```diff
@@ -98,6 +99,60 @@ def update_url_in_setup_py(package_dir: str):
...
+            client_specific_post_processing_path
+        ).iterdir():
+            with open(post_processing_path, "r") as post_processing_path_file:
+                post_processing_json = yaml.safe_load(post_processing_path_file)
+                all_replacements = post_processing_json["replacements"]
+                # For each workaround related to the specified issue
+                for replacement in all_replacements:
+                    # For each file that needs the workaround applied
+                    for client_library_path in replacement["paths"]:
+                        if package_name in client_library_path:
```
vchudnov-g, line 136: The semantics of this `if` are very targeted, but this isn't strictly needed since the replacements are all idempotent. But let's solve the issue on the assertion against `replacement["count"]` first (see other comment thread on the code below) before changing this.

tests/test_python_mono_repo.py
```diff
@@ -0,0 +1,51 @@
...
+            "packages/google-cloud-workflows/docs/index.rst"
+        ).read_text()
+        assert (
+            docs_index_rst
+            == "Completed replacement 1.\nCompleted replacement 1.\nCompleted replacement 2.\n"
+        )
+
+        # Confirm that `AssertionError`` is raised if you run the post processing more than once
+        # because the replacement should only occur once.
+        with pytest.raises(AssertionError):
```
vchudnov-g, line 47: I thought we said the script should be idempotent? I don't think this should raise.

(This issue is separate from but very related to the idempotency issue in the other thread.)

synthtool/languages/python_mono_repo.py
parthea: fixed in a9315ac

synthtool/languages/python_mono_repo.py
parthea: `replacement["paths"]` must only include paths for a single package so that `replacement["count"]` reflects only the replacement for a single package. I fixed the description in https://github.com/googleapis/synthtool/pull/1881/commits/04d32e919280e96cf1a7798e783cdaee44014687. Each package needs to have a separate `-replacement` section in the yaml.

synthtool/languages/python_mono_repo.py
parthea: I clarified the usage of `replacement["paths"]` in https://github.com/googleapis/synthtool/commit/04d32e919280e96cf1a7798e783cdaee44014687 to say that all of the paths must be for the same package.

tests/test_python_mono_repo.py
parthea: I think the script is idempotent because running the script more than once won't cause unintended replacements. You would get the same generated code, regardless of how many times you run the script.

To add more background, the client specific post processing runs selectively only when the generator runs (when the directory `owl-bot-staging/<package-name>` exists). This logic is already part of the python post processor here:

https://github.com/googleapis/synthtool/blob/d52e638b37b091054c869bfa6f5a9fedaba9e0dd/synthtool/languages/python_mono_repo.py#L145

The current behaviour is:

- There is a copy-code job which is part of the owlbot design that copies code from googleapis-gen and puts it into `owl-bot-staging/<package name>`
- We only run the post processor logic when the `owl-bot-staging/<package name exists>`
- The client specific post processing will only during the post processing logic. 

This assert is checking that if we try to run the client specific post processing again without an `owl-bot-staging` folder, there should be no changes to ensure that if the client specific post processing again there won't be unintended replacements.

synthtool/languages/python_mono_repo.py
parthea: Resolving old thread

synthtool/languages/python_mono_repo.py
parthea: Resolving old thread

synthtool/languages/python_mono_repo.py
vchudnov-g: OK, clarified offline. In essence, `all_replacements` is a list of the replacements _per package_ needed to implement this workaround, and each of these will now refer to a single package and have a count for verification. +1

tests/test_python_mono_repo.py
parthea: I added a comment in f92efba to clarify that the method `apply_client_specific_post_processing` will only be called when `owl-bot-staging` is present.

synthtool/languages/python_mono_repo.py
parthea: The current design already has a separate count per package. There will be a separate entry for each replacement under `replacements` See below:
```
replacements:
  - paths: ["packages/google-cloud-workflows/docs/index.rst"]
    before: "Second replacement"
    after: "Completed replacement 2."
    count: 1
  - paths: ["packages/some-other-package/docs/index.rst"]
    before: "Does not exist"
    after: "Replacement does not exist"
    count: 1
```

synthtool/languages/python_mono_repo.py
```diff
@@ -152,6 +153,11 @@ def apply_client_specific_post_processing(
                                 )
                                 == 0
                             )
+                    if validate_replacements:
```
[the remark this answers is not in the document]
parthea: No, because the count is for all paths, not a single path.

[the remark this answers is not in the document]
vchudnov-g: Oh, I see. You're checking for the total counts summed over multiple possible paths  for the same package. But it seems to me you still don't need `validate_replacements`, because you only go into the outer `if` if there are replacements, and `replacement_count` starts at 0 for each package.

vchudnov-g (approved):
Thanks for clarifying my question offline.

Please respond to my two comments about simplifying the code (the assertion against 0 and the non-need for `validate_replacements`), but LGTM.

synthtool/languages/python_mono_repo.py
```diff
@@ -98,6 +99,66 @@ def update_url_in_setup_py(package_dir: str):
...
+                        if package_name in client_library_path:
+                            validate_replacements = True
+                            replacement_count += synthtool.replace(
+                                client_library_path,
+                                replacement["before"],
+                                replacement["after"],
+                            )
+                            # Ensure idempotency by checking that subsequent calls won't
+                            # trigger additional replacements within the same path
+                            assert (
```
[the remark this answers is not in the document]
parthea: I don't think this check can be removed as it's making sure that the replacement is idempotent. It's asserting that multiple calls will not trigger additional replacements.

synthtool/languages/python_mono_repo.py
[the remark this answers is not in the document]
vchudnov-g: oh, I see: we could be called for a `package` where this workaround doesn't apply, so we don't expect any of these replacements to be applied. OK.

I wonder whether it makes sense to have a coherence check that all the `client_library_path`s in a single `replacement["paths"]` refer to the same package.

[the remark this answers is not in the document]
parthea: Fixed in [5f6b5f8](https://github.com/googleapis/synthtool/pull/1881/commits/5f6b5f82c97e8692e7b93d08caf146f8fca26268)

synthtool/languages/python_mono_repo.py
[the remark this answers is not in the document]
parthea: Removed in [5f6b5f8](https://github.com/googleapis/synthtool/pull/1881/commits/5f6b5f82c97e8692e7b93d08caf146f8fca26268) and replaced with `number_of_paths_with_replacements` to ensure that all paths are for the same package.

PR #1882, opened 2023-10-17
chore(python): in .coveragerc, ignore branches that aren't really branches

tswast:
We don't expect `with pytest.raises...:` "branches" to fail in passing tests.

### DIFF
--- a/synthtool/gcp/templates/python_library/.coveragerc
+++ b/synthtool/gcp/templates/python_library/.coveragerc
@@ -31,6 +31,8 @@ exclude_lines =
     def __repr__
     # Ignore abstract methods
     raise NotImplementedError
+    # Ignore branches that aren't really branches in tests
+    with pytest.raises
 omit =
   */gapic/*.py
   */proto/*.py


parthea:
@tswast , Please could you file an issue https://github.com/googleapis/synthtool/issues/new?assignees=&labels=&projects=&template=bug_report.md with a link to an example of the problem to help with the review?

tswast:
Strangely I'm not able to reproduce anymore. Maybe it was a bad "coverage" package release or something?

tswast:
For future reference, I was able to reproduce when I only ran the Python 3.9 unit tests, followed by cover. With Python 3.10 unit tests, the coverage report correctly omits these "branches".

parthea:
I'm not sure if it's related but I ran into a similar issue where there was a python 3.9 coverage gap when testing async code.

https://github.com/googleapis/gapic-generator-python/pull/1174#issuecomment-1025132372
https://github.com/nedbat/coveragepy/issues/1223

PR #1886, opened 2023-10-20
feat: add esm options for Nodejs

sofisl:
Most of the files are templates, the main changes are:

- Add an ESM path to generating index.ts (necessary since import paths are changing, and since the index.ts now lives one level lower than previously)
- Add an ESM option to generate protos in es6 (soon this will be obsolete, since we moved it to the generator, but this is necessary until we remove compileProtos from here)
- Add an ESM path for templates (mainly, changing .js --> .cjs)


sofisl:
@bcoe, would you mind taking a look whenever you have a minute? Thank you!

SurferJeffAtGoogle:
In the pull request description, could you describe *why* these files are changing?

bcoe (changes requested):
I believe the majority of the `node_esm_mono_repo_library` files are the same as the non-ESM versions?

Could we consider giving jinja multiple search paths:

```
loader = FileSystemLoader(["/override/templates", "/default/templates"])
```

And just add the template files that differ?

If this change could be done without too much refactoring, it would make it much easier to update templates.

bcoe (approved):
Approving, with the suggestion that we refactor to use two template directories, so that we don't need duplicate copies of CONTRIBUTING, LICENSE, .eslintignore, etc.

sofisl:
> Approving, with the suggestion that we refactor to use two template directories, so that we don't need duplicate copies of CONTRIBUTING, LICENSE, .eslintignore, etc.

After looking into this, I'm a bit hesitant to do this because it would require refactoring [this function](https://github.com/googleapis/synthtool/blob/1063ef32bfe41b112bade7a2dfad4e84d0058ebd/synthtool/gcp/common.py#L52) (which renders the templates) to accept a List of template paths, vs. a string. That would affect Python and Java as well. 


I've created a bug for this: https://github.com/googleapis/synthtool/issues/1897

PR #1891, opened 2023-11-02
chore: update docfx minimum Python version

dandhlee:
Sphinx plugin now requires Python3.10 as the minimum version, which is run in `docfx` sessions. This change is required to accommodate for the breaking change for upgrade in the version. See https://github.com/googleapis/sphinx-docfx-yaml/pull/335.

For split repos, only the `docfx` session is updated to run in 3.10, and will respect what the split repo maintainer chooses to use as the default version.

For the monorepo, 3.10 will be used as the default version in general. Both split and monorepo uses Ubuntu22.04, which has Python3.10 as the built in Python version.

Unblocks major release for https://github.com/googleapis/sphinx-docfx-yaml/pull/335.

### DIFF
--- a/synthtool/gcp/templates/python_library/.github/workflows/docs.yml
+++ b/synthtool/gcp/templates/python_library/.github/workflows/docs.yml
@@ -28,7 +28,7 @@ jobs:
     - name: Setup Python
       uses: actions/setup-python@v4
       with:
-        python-version: "3.9"
+        python-version: "3.10"
     - name: Install nox
       run: |
         python -m pip install --upgrade setuptools pip wheel
--- a/synthtool/gcp/templates/python_library/noxfile.py.j2
+++ b/synthtool/gcp/templates/python_library/noxfile.py.j2
@@ -370,7 +370,7 @@ def docs(session):
     )
 
 
-@nox.session(python="3.9")
+@nox.session(python="3.10")
 def docfx(session):
     """Build the docfx yaml files for this library."""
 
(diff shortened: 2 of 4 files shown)


parthea:
@dandhlee  I tried creating a similar PR in google-cloud-python but ran into errors. Please could you check the changes made in PR https://github.com/googleapis/google-cloud-python/pull/11978?

```
Could not import extension docfx_yaml.extension (exception: cannot import name 'Callable' from 'collections' (/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/collections/__init__.py))
```

```
ImportError: cannot import name 'Union' from 'types' (/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/types.py)
```

parthea:
Updating sphinx to 4.5.0 fixes the error `ImportError: cannot import name 'Union' from 'types' `

https://github.com/googleapis/synthtool/blob/1547f9a33de075d31d434bb3c9d7f53f068e0ff3/synthtool/gcp/templates/python_mono_repo_library/noxfile.py.j2#L362

parthea:
`Could not import extension docfx_yaml.extension (exception: cannot import name 'Callable' from 'collections' (/opt/hostedtoolcache/Python/3.10.13/x64/lib/python3.10/collections/__init__.py))` will be resolved once https://github.com/googleapis/sphinx-docfx-yaml/pull/335 is merged.

parthea (commented):
LGTM. Only one observation that the minimum version of sphinx should be updated in the docs build from 4.0.1 to 4.5.0

dandhlee:
Done! Bumped monorepo to 4.5 - I suppose this is because the default is now Python3.10 for both. I'll leave it for split repos because it has caused issues for some libraries before and docs will still work with Python3.9. Please take a look again.

parthea (approved):

PR #1900, opened 2023-11-15
feat: add filepath to esm index.ts file

sofisl:
ESM needs the full correct filepath to find the file (vs. CJS, which just needed the directory)


alexander-fenster (approved):

bcoe (approved):
I'd probably call this a `feat` or ` fix`, since it's making a functional change to the generated libraries.

PR #1910, opened 2023-12-14
build: update actions/upload-artifact and actions/download-artifact

parthea:
https://github.com/actions/upload-artifact/releases/tag/v4.0.0
https://github.com/actions/download-artifact/releases/tag/v4.0.0

There are breaking changes in `upload-artifact` v4 where the artifact name must be different. See https://github.com/actions/toolkit/tree/main/packages/artifact#breaking-changes

```
Due to how Artifacts are created in this new version, it is no longer possible to upload to the same named Artifact multiple times. You must either split the uploads into multiple Artifacts with different names, or only upload once.
```

I tested the changes in https://github.com/googleapis/python-pubsublite/pull/466


vchudnov-g (approved):
Thanks for dealing with this!

PR #1917, opened 2024-01-18
chore: update java template license header year

diegomarquezp:


suztomo (approved):

synthtool/languages/java.py
```diff
@@ -29,7 +29,7 @@
 JAR_DOWNLOAD_URL = "https://github.com/google/google-java-format/releases/download/google-java-format-{version}/google-java-format-{version}-all-deps.jar"
 DEFAULT_FORMAT_VERSION = "1.7"
 GOOD_LICENSE = """/*
- * Copyright 2023 Google LLC
```
blakeli0, line 32: Can we make it a variable so it automatically updates to current year? The tests would automatically start failing in a new year, but I think it could be a better way to force us update the copyright in time. WDYT @suztomo ?

suztomo: That's a good idea.

diegomarquezp: I added a variable to compute the license year.

blakeli0 (approved):

bcoe (approved):

PR #1919, opened 2024-01-19
chore: remove unused templates in java_library

JoeWang1127:
In this PR:
- Remove template files in samples/ directory in java library.


suztomo:
I see it's used https://github.com/googleapis/java-bigtable/blob/main/samples/install-without-bom/pom.xml

suztomo (changes requested):
Joe to think about what to do with the template which is unused in the monorepo but used in the handwritten repositories.

PR #1939, opened 2024-03-15
chore(python): add requirements for docs build

parthea:
This PR fixes an issue where the `Kokoro docs-presubmit` job fails with `/usr/bin/python3: No module named nox`. See PR https://github.com/googleapis/python-access-context-manager/pull/206 where the issue was seen and the issue was fixed.


shollyman (approved):
Thanks for putting this together!

holtskinner (approved):

PR #1945, opened 2024-03-27
chore: update ignore paths in renovate config

JoeWang1127:
In this PR:
- Update renovate config to explicitly ignore workflow files in `.github/workflows`.

Context: we want to use renovateBot to update the version of unmanaged dependency check in `.github/workflows/unmanaged_dependency_check.yaml`. However, the directory is ignored in renovate config (#1855). Explicitly list files in this directory can make renovate include `.github/workflows/unmanaged_dependency_check.yaml`.

### DIFF
--- a/synthtool/gcp/templates/java_library/renovate.json
+++ b/synthtool/gcp/templates/java_library/renovate.json
@@ -14,7 +14,10 @@
   ],
   "ignorePaths": [
     ".kokoro/requirements.txt",
-    ".github/workflows/**"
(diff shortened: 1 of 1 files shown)


suztomo (approved):

blakeli0:
I think we have to port this change to the [templates in sdk-platform-java](https://github.com/googleapis/sdk-platform-java/blob/main/library_generation/owlbot/templates/java_library/renovate.json) as well?

PR #1956, opened 2024-04-15
feat: add `libraries_bom_version` in metadata

JoeWang1127:
In this PR:
- Add a `libraries_bom_version` when rending `README.md`. It will fall back to `latest_bom_version` if this parameter is not set.


JoeWang1127:
These lines prevent me to write a unit test running in kokoro.

```
# skip README generation on Kokoro (autosynth)
    if os.environ.get("KOKORO_ROOT") is not None:
        # README.md is now synthesized separately. This prevents synthtool from deleting the
        # README as it's no longer generated here.
        excludes.append("README.md")
```

blakeli0 (approved):

PR #1969, opened 2024-05-27
chore(python): allow post processor to run without owlbot copy-code

parthea:
There are a few use cases where it would be helpful to be able to run the post processor in the python monorepo on an ad-hoc basis. We can do this today by following these steps:

- Clone https://github.com/googleapis/google-cloud-python
- Run `mkdir -p owl-bot-staging/<some-package>`
- Run the post processor `docker run --user $(id -u):$(id -g) --rm -v $(pwd):/repo -w /repo gcr.io/cloud-devrel-public-resources/owlbot-python-mono-repo:latest`

After completing these steps, the post processor runs successfully however the samples directory is erroneously deleted. The reason is that we cleaned up the samples directory but did not create new samples because skipped the step where new clients were generated. 

This fixes the issue where samples are erroneously deleted when running the post processor on an ad-hoc basis.


parthea:
@vchudnov-g @ohmayr PTAL

ohmayr (approved):

PR #1979, opened 2024-06-25
feat: update node library question template

leahecole:
Updates the "Question" template for node library issues to also ask for Node version, npm version, library version


leahecole:
I think this is obsolete now with the amazing work @sofisl has done so I'm closing!

PR #1988, opened 2024-07-08
chore(node): use NODE_ENV instead of flags to only install prod deps

ddelgrosso1:


ddelgrosso1:
The `--production` flag that was being used is not up to date with more recent versions of `npm`. For example npm 6.x uses `--only` and npm 8.x uses `--omit`. Probably better to set NODE_ENV to production instead of trying to keep up with the flags.

PR #2005, opened 2024-08-15
chore(python): Include docfx in nox.options.sessions

parthea:


synthtool/gcp/templates/python_library/noxfile.py.j2
```diff
@@ -139,6 +138,7 @@ nox.options.sessions = [
     "lint_setup_py",
     "blacken",
     "docs",
+    "docfx",
```
ohmayr, line 141: Question: What changed that it wasn't needed before but is now?

ohmayr (approved):

synthtool/gcp/templates/python_library/noxfile.py.j2
parthea: Some repositories had a separate docfx Kokoro job but we're slowly removing that presubmit in favor of running `docfx` via the `Kokoro` job or via `Github Actions`

PR #2008, opened 2024-09-04
chore(python): update unittest workflow template

chelsea-lin:
This change is adding `include-hidden-files` to the `upload-artifact` step. The `upload-artifact` excludes hidden files by default [here](https://github.com/actions/upload-artifact/commit/50769540e7f4bd5e21e526ee35c689e35e0d6874) so that `.coverage-*` files would be excluded to be uploaded.


parthea (approved):
Thanks @chelsea-lin!

PR #2011, opened 2024-09-18
feat!: Support Node 18+

danielbankhead:
Adding support for Node 18+. Dropping older versions of Node.


tests/fixtures/nodejs_mono_repo_without_version/packages/no_version/package.json
```diff
@@ -4,7 +4,7 @@
   "license": "Apache-2.0",
   "repository": "googleapis/repo-automation-bots",
   "engines": {
```
sofisl, line 6: I think we should keep the tests and fixtures as is, these are unnecessary changes.

synthtool/gcp/templates/node_library/.kokoro/test.bat
```diff
@@ -21,7 +21,7 @@ cd ..
 @rem we upgrade Node.js in the image:
 SET PATH=%PATH%;/cygdrive/c/Program Files/nodejs/npm
 
-call nvm use v14.17.3
+call nvm use 18
```
sofisl, line 24: should it be 18.17? https://nodejs.org/en/blog/release/v18.17.0

synthtool/gcp/templates/node_library/.kokoro/release/publish.cfg
```diff
@@ -30,7 +30,7 @@ build_file: "{{ metadata['repository_name'] }}/.kokoro/trampoline_v2.sh"
 # Configure the docker image for kokoro-trampoline.
 env_vars: {
     key: "TRAMPOLINE_IMAGE"
-    value: "gcr.io/cloud-devrel-kokoro-resources/node:14-user"
+    value: "gcr.io/cloud-devrel-kokoro-resources/node:18-user"
```
sofisl, line 33: I'm sure this should exist, because: https://github.com/googleapis/testing-infra-docker/blob/main/node/18-user/Dockerfile

but I also don't see it here: https://pantheon.corp.google.com/artifacts/docker/cloud-devrel-kokoro-resources/us/gcr.io?inv=1&invt=AbcqFw&project=cloud-devrel-kokoro-resources

would you mind confirming this image is usable?

synthtool/gcp/templates/node_library/.kokoro/presubmit/node14/common.cfg
```diff
@@ -16,7 +16,7 @@ build_file: "{{ metadata['repository_name'] }}/.kokoro/trampoline_v2.sh"
 # Configure the docker image for kokoro-trampoline.
```
sofisl, line 16: these files and folders should also be renamed to node 18

sofisl (changes requested):

tests/fixtures/nodejs_mono_repo_without_version/packages/no_version/package.json
sofisl: This goes for all changes under the tests/ directory, it bloats the PR.

sofisl:
We'll want to follow up with a PR to update the Docker image dependencies once these are updated: https://github.com/googleapis/synthtool/blob/master/docker/owlbot/nodejs_mono_repo/Dockerfile#L55C1-L56C79

synthtool/gcp/templates/node_library/.kokoro/test.bat
danielbankhead: This should install and use the latest version of Node 18

sofisl: I think it's better to specify given we had that before, is there a reason you think it shouldn't be specified?

sofisl (changes requested):

synthtool/gcp/templates/node_library/.kokoro/test.bat
danielbankhead: This would ensure the version tested has all of the latest bug fixes and features.

danielbankhead: I see we have a pattern from previous iterations - however it doesn't necessarily mean it was best practice (even that linked PR suggests the same - the patch was a required manual update that we shouldn't have to do each time).

The stability for the oldest supported LTS is far less-likely to receive a bug or breaking change than a newer LTS, while taking advantage of any provided security patches. See the following:
- https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V18.md
- https://github.com/nodejs/node/blob/main/doc/changelogs/CHANGELOG_V14.md

In short, we'd be preferring the problem of slight risk of a broken release vs getting pinged for manual security patching. I'd rather deal with the slight risk of a broken release as it would block releases across the ecosystem (all open source Node projects), which wouldn't make it a pressing matter for just us.

sofisl (approved):

PR #2013, opened 2024-09-20
chore: remove java owlbot

JoeWang1127:
In this PR:
- Rmove Java Owlbot since the source code is migrated to sdk-platform-java and it is disabled in non-GAPIC libraries (see go/disable-owlbot-java).


tests/test_language_java.py
blakeli0, line 1: I don't see this test being migrated to sdk-platform-java and looks like it still provides value, can you try to move it?

docker/owlbot/java/cloudbuild.yaml
blakeli0, line 1: I assume there are some cloud build triggers associated with these cloud build yamls? Can we figure out what are them and delete them as well?

JoeWang1127: I created b/368584265 so @chingor13 can remove it.

tests/test_language_java.py
JoeWang1127: Traced in https://github.com/googleapis/sdk-platform-java/pull/3227

blakeli0 (approved):

chingor13 (approved):

PR #2050, opened 2025-01-10
chore(python_mono_repo): update template to cater for grpc-google-iam-v1

parthea:
Currently we have a check in an if statement for whether a package has `google` in the name to determine if we should lint the folder `google`. The specific check is `metadata['repo']['distribution_name'].startswith('google')`. This doesn't work for the [grpc-google-iam-v1](https://pypi.org/project/grpc-google-iam-v1/) package. We should use `'google' in metadata['repo']['distribution_name']` instead.

### DIFF
--- a/synthtool/gcp/templates/python_mono_repo_library/noxfile.py.j2
+++ b/synthtool/gcp/templates/python_mono_repo_library/noxfile.py.j2
@@ -28,7 +28,7 @@ import nox
 
 BLACK_VERSION = "black[jupyter]==23.7.0"
 ISORT_VERSION = "isort==5.11.0"
-{% if metadata['repo']['distribution_name'].startswith('google') %}
+{% if 'google' in metadata['repo']['distribution_name'] %}
 LINT_PATHS = ["docs", "google", "tests", "noxfile.py", "setup.py"]
 {% else %}
 LINT_PATHS = ["docs", "{{ metadata['repo']['distribution_name'] }}", "tests", "noxfile.py", "setup.py"]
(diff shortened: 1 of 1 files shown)


synthtool/gcp/templates/python_mono_repo_library/noxfile.py.j2
```diff
@@ -28,7 +28,7 @@ import nox
 
 BLACK_VERSION = "black[jupyter]==23.7.0"
 ISORT_VERSION = "isort==5.11.0"
-{% if metadata['repo']['distribution_name'].startswith('google') %}
+{% if 'google' in metadata['repo']['distribution_name'] %}
```
ohmayr, line 31: What about `googleapis-common-protos` that we recently added and `grafeas`. Did we not have a similar issue for these?

ohmayr (approved):

synthtool/gcp/templates/python_mono_repo_library/noxfile.py.j2
parthea: `googleapis-common-protos` is fine because the check `'google' in metadata['repo']['distribution_name']` passes and we currently lint the `google` directory. This line is added to templates `LINT_PATHS = ["docs", "google", "tests", "noxfile.py", "setup.py"]`. For `grafeas`, the check does not pass, so we end up with `LINT_PATHS = ["docs", "{{ metadata['repo']['distribution_name'] }}", "tests", "noxfile.py", "setup.py"]`, which generates as `LINT_PATHS = ["docs", "grafeas", "tests", "noxfile.py", "setup.py"]`. IOW, both `googleapis-common-protos`  and `grafeas` are correct.

See the nox files here:

grafeas:
https://github.com/googleapis/google-cloud-python/blob/3a9a8fb2be1304ff8d4593320236e1ea008ee696/packages/grafeas/noxfile.py#L33

`googleapis-common-protos`
https://github.com/googleapis/google-cloud-python/blob/3a9a8fb2be1304ff8d4593320236e1ea008ee696/packages/googleapis-common-protos/noxfile.py#L33

These nox files will remain unchanged after this PR is merged.

PR #2052, opened 2025-01-15
build: add tests to template automation and clean up templating logic

sofisl:
Fixes a bug that closed issues prematurely due to formatting. Also adds tests so that library maintainers can update their own scripts if needed and know if they are breaking any automation.


danieljbruce (approved):
LGTM. I don't know too much about the synthtool though.

leahecole (approved):

PR #2053, opened 2025-01-15
chore: use Airlock for releasing

sofisl:


synthtool/gcp/templates/node_library/.kokoro/continuous/node18/common.cfg
```diff
@@ -21,4 +21,4 @@ env_vars: {
 env_vars: {
     key: "TRAMPOLINE_BUILD_FILE"
     value: "github/{{ metadata['repository_name'] }}/.kokoro/test.sh"
-}
```
sakthivelmanii, line 24: Can we remove the change from this file?

synthtool/gcp/templates/node_library/.kokoro/trampoline_v2.sh
```diff
@@ -165,6 +165,7 @@ if [[ -n "${KOKORO_BUILD_ID:-}" ]]; then
 	# For flakybot
 	"KOKORO_GITHUB_COMMIT_URL"
 	"KOKORO_GITHUB_PULL_REQUEST_URL"
+    "USE_AIRLOCK"
```
sakthivelmanii, line 168: Can you please put tab instead of space here? I copied from the previous line

sakthivelmanii (approved):

synthtool/gcp/templates/node_library/.kokoro/presubmit/node18/system-test.cfg
```diff
@@ -9,4 +9,9 @@ env_vars: {
 env_vars: {
   key: "SECRET_MANAGER_KEYS"
   value: "long-door-651-kokoro-system-test-service-account"
+}
+
+env_vars: {
```
sakthivelmanii, line 14: There are two ways to let each library enable it. Each team have their own set of dependencies and requires a follow up with Airlock team to make sure their dependencies are available.

1. By default disable it and once synthool changes are merged in each library, each client can enable it.(Flipping the flag to true).(Defining something like {{ false or true} } not sure if it's possible)
2. Enable it by default(set it to true). So that before merging the PR, each library can make sure all their dependencies available before merging synthool PR. Drawback is that subsequent synthool might be blocked.

sakthivelmanii: For Spanner, I identified the all the dependencies which are missing in Airlock using a script and created a ticket for Airlock team to import those dependencies. I will share the script over DM.

All the libraries will work with Airlock but we can't confirm that all the dependencies will be available in Airlock. For missing dependencies, library owner have to reach out to Airlock team.

sofisl: Great, feel free to email it to me if that's better. Thanks!

suztomo:
I got an idea of creating a new "Build with Airlock" check rather than reusing an existing one. It's applied to a Java repository and looks like this:

image

More in b/356853775.

sakthivelmanii:
@suztomo That's good idea. We can follow that in Node as well.

PR #2069, opened 2025-03-10
chore(python): Update renovate.json to exclude pyproject.toml

parthea:
See https://github.com/googleapis/python-bigquery/pull/2143 where renovate bot is proposing undesired changes to `pyproject.toml`. 

We're currently excluding `setup.py` but we also need to exclude `pyproject.toml`

https://github.com/googleapis/synthtool/blob/bb0a3506602525f63c7002f8d13345be3678effb/synthtool/gcp/templates/python_library/renovate.json#L8

This PR also includes the following fixes:
- exclude `.kokoro/requirements.txt`  which was removed in https://github.com/googleapis/synthtool/pull/2035

### DIFF
--- a/synthtool/gcp/templates/python_library/renovate.json
+++ b/synthtool/gcp/templates/python_library/renovate.json
@@ -5,7 +5,7 @@
     ":preserveSemverRanges",
     ":disableDependencyDashboard"
   ],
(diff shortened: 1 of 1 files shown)


parthea:
Adding `do not merge` as I still need to include changes from https://github.com/googleapis/synthtool/pull/2071 but I will wait until after 2071 is merged

PR #2082, opened 2025-04-09
chore(python): remove MANIFEST.in from templates

parthea:
Remove `MANIFEST.in` in synthtool templates. This won't delete existing `MANIFEST.in` files downstream. The downstream files must not be removed. This PR only removes the file from templates so that downstream files are maintained as handwritten ones.


ohmayr (approved):

parthea:
Wait for https://github.com/googleapis/gapic-generator-python/pull/2367 to be released

PR #2093, opened 2025-05-28
chore: update NodeJS post-processor to use MOSS approved images and update node version for tests

sofisl:


docker/owlbot/nodejs/Dockerfile
```diff
@@ -16,39 +16,92 @@
...
+    libsqlite3-dev \
+    wget \
+    libbz2-dev \
+    tk-dev \
+    liblzma-dev \
+    curl \
+    git && \
+    rm -rf /var/lib/apt/lists/*
+
+# Define the Python version for easier updates and download/install Python from source.
```
meltsufin, line 47: Why did you chose to install Python from source rather than with `apt-get`?

sofisl: Changed.

meltsufin: Thanks!

docker/owlbot/nodejs/Dockerfile
```diff
@@ -12,43 +12,81 @@
...
+
+## Install Node.js
 
-###################### Install nodejs.
-RUN curl https://nodejs.org/dist/v18.20.8/node-v18.20.8-linux-x64.tar.xz > /tmp/nodejs.tar.xz
-RUN tar -C /usr/local --strip-components=1 -xJf /tmp/nodejs.tar.xz
-RUN rm -f /tmp/nodejs.tar.xz
+# Download and extract Node.js directly into /usr/local.
+# The 'curl' command includes '-fsSL' for silent, fail-fast, and redirect-following behavior.
+RUN curl -fsSL https://nodejs.org/dist/v18.20.8/node-v18.20.8-linux-x64.tar.xz -o /tmp/nodejs.tar.xz && \
```
meltsufin, line 47: Can Node come from apt-get as well?

sofisl: Is there a security reason for it? In this case we're changing what was previously in the Dockerfile so I'd rather keep it is as unless it's better to do so with apt-get.

meltsufin (dismissed):

docker/owlbot/nodejs/Dockerfile
sofisl: ok great, I updated!

meltsufin (approved):

alvarowolfx (approved):

PR #2116, opened 2025-08-15
feat: add code for new post processor

sofisl:
see https://docs.google.com/document/d/1hxAtt0FaMp6vZX3Z57E62qqzvzsP3EzRE56N4ye7fzo/edit?tab=t.0#heading=h.xzptrog8pyxf

and https://github.com/googleapis/google-cloud-node/pull/6604
and https://github.com/googleapis/gapic-generator-typescript/pull/1811

This change to synthtool simply adds a new path to handle this new library.

We are not adding tests for this functionality for two reasons:

1) We are deprecating synthtool soon so it's not worth it.
2) Most of the functionality we're adding is just a script that runs the CLI tool that is well unit tested (gapic-node-processing)


feywind (dismissed):
It's exciting that it looks like stuff is moving out of synthtool.

chingor13 (approved):

PR #2127, opened 2025-09-30
fix: disable renovate for Node github action YAML configs

leahecole:
We have templated GitHub action configs and renovate opens PRs to update them, which then get clobbered by owlbot shortly after - this I think fixes b/444433904


sofisl (approved):
Nit: let's create an issue for this since the gh action file won't get an update anynore

PR #2131, opened 2025-10-09
chore(python): Add Python 3.14 to python post processor image

parthea:
Towards b/375664027


abbrowne126 (dismissed):
Two suggestions about capitalization but I see [this is consistent with previous versions](https://github.com/googleapis/synthtool/blob/796239424329a8e36263b96d6d57db2eede540ed/synthtool/gcp/templates/python_library/.kokoro/samples/python3.13/periodic.cfg#L11) so it may not apply

abbrowne126 (approved):

PR #2180, opened 2026-08-06
feat: add git concurrency patch

sofisl:
## Description
This PR introduces an fcntl.flock file lock around the git clone, git checkout, and git pull operations within synthtool.sources.git.clone.

Context & Motivation: As part of the transition away from using Docker for librarian generate on Python packages (via the removal of librarian.yaml), client library generation is now executed locally on the host machine.

When running librarian generate -v --all, generation tasks are executed concurrently across multiple packages. This concurrency introduces a race condition where multiple processes attempt to clone or pull the same shared repository in the synthtool cache simultaneously, resulting in git index.lock collision errors that fail the generation.

By applying an exclusive file lock (.lock) over the specific cache repository before manipulating it, we guarantee that concurrent librarian processes serialize their git operations safely and avoid index.lock collisions.

## Changes Made
Added an fcntl.flock block in synthtool/sources/git.py that surrounds the git clone, git checkout, and git pull subprocess calls.
Added testCloneConcurrencyPatch to tests/test_git.py using mock.patch to verify that the LOCK_EX and LOCK_UN calls are correctly executed during cloning.
Corrected the indentation block for the try/finally block to ensure synthtool parses it properly.

### DIFF
--- a/synthtool/sources/git.py
+++ b/synthtool/sources/git.py
@@ -105,15 +105,16 @@ def clone(
 
         dest = dest / pathlib.Path(url).stem
 
-        if force and dest.exists():
-            shutil.rmtree(dest)
+    import fcntl
+    lock_file = dest.parent / (dest.name + ".lock")
+    with open(lock_file, "w") as lock_f:
+        fcntl.flock(lock_f, fcntl.LOCK_EX)
+        try:
+            if not preclone:
+                if force and dest.exists():
+                    shutil.rmtree(dest)
 
-        default_branch = None
-        import fcntl
-        lock_file = dest.parent / (dest.name + ".lock")
-        with open(lock_file, "w") as lock_f:
-            fcntl.flock(lock_f, fcntl.LOCK_EX)
-            try:
+                default_branch = None
                 if not dest.exists():
                     cmd = ["git", "clone", "--recurse-submodules", "--single-branch", url, dest]
                     shell.run(cmd, check=True)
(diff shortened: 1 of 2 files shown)


synthtool/sources/git.py
```diff
@@ -105,21 +105,39 @@ def clone(
...
-            shell.run(cmd, check=True)
-        else:
-            default_branch = _local_default_branch(dest)
-            shell.run(["git", "checkout", default_branch], cwd=str(dest), check=True)
-            shell.run(["git", "pull"], cwd=str(dest), check=True)
-        committish = committish or default_branch
-
-    if committish:
-        shell.run(["git", "reset", "--hard", committish], cwd=str(dest))
+    import fcntl
```
daniel-sanche: can we keep imports at the top?

synthtool/sources/git.py
```diff
@@ -105,21 +105,39 @@ def clone(
...
+                    shell.run(
+                        ["git", "checkout", default_branch], cwd=str(dest), check=True
+                    )
+                    shell.run(["git", "pull"], cwd=str(dest), check=True)
+                committish = committish or default_branch
+
+            if committish:
+                shell.run(["git", "reset", "--hard", committish], cwd=str(dest))
+        finally:
+            fcntl.flock(lock_f, fcntl.LOCK_UN)
```
daniel-sanche: suggestion: it might be cleaner to extract the lock resource into a separate context manager helper:

```
@contextlib.contextmanager
def file_lock(lock_path: pathlib.Path):
    with open(lock_path, "w") as lock_f:
        fcntl.flock(lock_f, fcntl.LOCK_EX)
        try:
            yield lock_f
        finally:
            fcntl.flock(lock_f, fcntl.LOCK_UN)
```

Then you could just use a `with file_lock()` block, and keep the rest of this function more focused

PR #2182, opened 2026-09-15
fix: update the synth tool to match node version 22

danieljbruce:
## Description

Apiary was upgraded to Node v22 so we had to change the testing infrastructure in apiary to match Node 22 as well. However, the owlbot post-processor is still reverting changes to use the Node 18 test infrastructure as well as shown in https://github.com/googleapis/google-api-nodejs-client/pull/3999 so we have to update the synthtool here to make sure the post-processor writes changes that are Node v22 compatible.

## Impact

Ensures post-processor doesn't make the wrong changes on the apiary PR.


feywind (approved):

danieljbruce:
It looks like this isn't completely necessary anymore because we have https://github.com/googleapis/google-api-nodejs-client/pull/4019 to turn the owlbot off, but I'll leave it here just in case.

